PR #8, opened 2019-08-26
fixed height map and visibility map

RaphaelRoyerRivard:
I'm not sure if other maps are using the same `SampleImage::GetBit` method, but since Timo mirrored maps that were starting at 0 on the top left, I guess that's better like this now.
Also, RobbyG is not getting the same terrain height values as I get, so we need someone else to test if the formula works because it does for me but not for him.


alkurbatov:
Thank you!

alkurbatov:
>I'm not sure if other maps are using the same SampleImage::GetBit method

This version of `GetBit` is used only in the heights and visibility maps.

PR #12, opened 2019-09-12
Fix: Pulled down update to protobuf to fix vs2019 compile error

Cryptyc:


alkurbatov:
The broken linux CI might be because we use ubuntu trusty as testing env (release 5 years ago). I believe it would be safe to switch to xenial.

alkurbatov:
Fixed in #36 Thank you for the idea!

PR #30, opened 2019-09-18
activate raw_affects_selection on release

RaphaelRoyerRivard:
To allow users to freely select units when debugging, it is activated only on Release.


RaphaelRoyerRivard:
Oh damn, I forgot about all these things :p
Passing the value as a parameter could be a good solution, I have not played with configuration of the API on the client side yet though.

alkurbatov:
Was reimplemented in #38

PR #31, opened 2019-09-19
Improve Travis-ci build and deploy steps

yanntm:
Added support for doxygen, simplified travis install procedure.

* solves issue with doxygen not building, travis has /build/ in it's work path
* drop the scripts in travis/, in favor of using additional env configuration in the .travis.yml itself
* deploy the generated documentation to github-pages : on successful build of Debug on linux on master branch, Doxygen is called and generated doc is uploaded to gh-pages


yanntm:
Note that for the gh-pages deployment to work you need to define a 
`GITHUB_TOKEN` in the settings of the travis, using the https://docs.travis-ci.com/user/deployment/pages/ "Repository Settings" approach

yanntm:
you can see the generated doc page on my fork : https://yanntm.github.io/cpp-sc2/

Also, as side benefit, anyone forking the project (and that defines the GH token) inherits the build of doc + gh-pages update, so maybe some text explaining this could be added to the part about editing the documentation in the main project Readme

yanntm:
sorry for 20 commits only changing 5 lines, I need to commit to trigger travis, the round-trip to debug is pretty long, and I spent some time before understanding the */build/* EXCLUDE issue.

alkurbatov:
> sorry for 20 commits only changing 5 lines, I need to commit to trigger travis, the round-trip to debug is pretty long, and I spent some time before understanding the _/build/_ EXCLUDE issue.

Many commits is not a problem. Please, do interactive rebase, squash the commits and do force push. Also, please add a tag to the commit message (e.g. `build:`) in order to make the changelog collector pick your work.

.travis.yml
```diff
@@ -7,32 +7,41 @@ matrix:
...
       env: BUILD_CONFIGURATION=Debug
     - os: osx
       osx_image: xcode9.1
       env: BUILD_CONFIGURATION=Release
 
-before_install:
-  - . .travis/before-install.sh
-
-install:
-  - . .travis/install.sh
```
alkurbatov: If we don't need those scripts anymore, it should be removed.

.travis.yml
```diff
@@ -7,32 +7,41 @@ matrix:
...
-      dist: trusty
-      group: travis_lts
-      sudo: required
-      env: BUILD_CONFIGURATION=Debug
+      compiler: gcc
+      addons: &gcc9
+        apt:
+          sources: ['ubuntu-toolchain-r-test']
+          packages: ['g++-9', 'gcc-9']
+      env: GCCVER='9' CXX_COMPILER=g++-${GCCVER} CXX=${CXX_COMPILER} C_COMPILER=gcc-${GCCVER} CC=${C_COMPILER} AR=gcc-ar-${GCCVER} RANLIB=gcc-ranlib-${GCCVER} NM=gcc-nm-${GCCVER} BUILD_CONFIGURATION=Release
```
alkurbatov: Please, move `BUILD_CONFIGURATION=Release` at the very beginning, just to emphasize which kind of build it is I think it is convenient.

.travis.yml
```diff
@@ -7,32 +7,41 @@ matrix:
...
-  - . .travis/before-install.sh
-
-install:
-  - . .travis/install.sh
-
 script:
   - mkdir build
   - cd build
   - . ../.travis/build.sh
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cd .. ; mkdir -p docs/html ; doxygen 2>&1 > doxygen.log || cat doxygen.log ; mkdir website ; cp -r docs/* website/ ; fi
```
alkurbatov: What do you think about moving this line into `build.sh`?

alkurbatov: Another question: is it possible to check the branch name here? Cause this code will call docs generation in every branch on linux twice (debug && release)

alkurbatov:
Shouldn't we manually specify which branches should be affected (e.g. master and dev)? As we have a separate gh-pages branch which could trigger builds.

.travis.yml
yanntm: yes

.travis.yml
yanntm: ok

yanntm:
ok I'll do the changes "soon" (TM)


> Shouldn't we manually specify which branches should be affected (e.g. master and dev)? As we have a separate gh-pages branch which could trigger builds.

currently we only deploy docs if the successful build was on the master branch, and only the linux/Debug config does the deploy
```
on:
    branch: master
    condition: $TRAVIS_OS_NAME = linux && $BUILD_CONFIGURATION = Debug
```

.travis.yml
alkurbatov: build.sh could be put into the `.travis` folder without any guarantee that it works somewhere else. But ok, lets keep the current approach.

>we can add a test to only generate in one of the cases
It would be good to do :)

I believe `generate_and_deploy_doxygen.sh` can be removed too?

.travis.yml
```diff
@@ -7,32 +7,41 @@ matrix:
...
-      sudo: required
-      env: BUILD_CONFIGURATION=Debug
+      compiler: gcc
+      addons: &gcc9
+        apt:
+          sources: ['ubuntu-toolchain-r-test']
+          packages: ['g++-9', 'gcc-9']
+      env: GCCVER='9' CXX_COMPILER=g++-${GCCVER} CXX=${CXX_COMPILER} C_COMPILER=gcc-${GCCVER} CC=${C_COMPILER} AR=gcc-ar-${GCCVER} RANLIB=gcc-ranlib-${GCCVER} NM=gcc-nm-${GCCVER} BUILD_CONFIGURATION=Release
     - os: linux
-      dist: trusty
```
alkurbatov: I think we should keep the `dist` name here. Of course Travis will pick some defaults for us but it would be good to explicitly specify which environment we use. Also, if Travis change the default bistro we will not get suddenly broken builds.

yanntm: my current best practice is to not specify the image for that reason, the default image is always a "stable" oldish one that is similar to what you can expect clients to have, but not *too* old. A breaking build on travis is not really an issue, it won't deploy, we will just know that we are using something that is now obsolete. We can set travis to build every month or so regardless of commits, so this kind of makes sure that we still run on something that is not obsolete. But the argument for "my travis didn't change, how did my build break" is also valid. 
So, as you prefer, we can put it back in, or even add a build without image specified (so that tests what happens when distros evolve), and one with "xenial" which is current default https://docs.travis-ci.com/user/reference/overview/
Anyway I'll follow whatever guideline you want, but personally I like the image-less version because it tests for obsolescence when travis update their setup (every year roughly). 
Like the current specified image was trusty, which is considered obsolete since april 2019

alkurbatov: I don't expect that something will be broken on the Travis side, but our build could fail because the environment has been changed silently. Also, it always gives a clear answer against which environment we validating.

>Like the current specified image was trusty, which is considered obsolete since April 2019

Thats ok and mostly indicate the lack of attention to the repo. Also, it explicitly guarantees that at least on trusty it works.

alkurbatov: To sum up: I suggest to set denial explicitly.

yanntm: ok

yanntm:
hi so I tried to implement the changes, I think the new variant for the build matrix is actually simpler. I also moved the definition of g++ -> g++-9 to a first line of script session, so it is less polluting to read the config. 
If you like this version I'll squash the commits

yanntm:
just looking at stats +35 -180 lines for added functionality is pretty cool :D

alkurbatov:
Oo that's really nice! :)
Please, squash the commit and will go with this version.

yanntm:
sorry the interactive rebase/squash thing did not go so well :D I don't usually attempt to rewrite history.
So let's close this one and use the clean version I built in another fork : https://github.com/alkurbatov/cpp-sc2/pull/35

alkurbatov:
Ok, lets move there.

PR #33, opened 2019-09-25
On neutral unit created

mboedigh:
I'm new to this and kind of did this PR the wrong way. I meant to have 3 independent PRs, but instead made them cumulative. 
This PR has 3 features
1) Caches Units experiencing certain events. This was for efficiency and clarity. This is the largest change and is a prerequisite for the other changes
2) Added OnUnitDamaged event, that gets issued when health or shields are weakened compared to the previous observation
3) Added OnNeutralUnitCreated. Effectively the equivalent of OnUnitCreated or OnUnitEntersVision, but for neutral units. Needed because prior to observing a neutral unit, it is a distinct placeholder unit that other units cannot interact with, and it is helpful to know when we can interact with them. 
Let me know if I should break this up into separate requests

### DIFF
--- a/include/sc2api/sc2_client.h
+++ b/include/sc2api/sc2_client.h
@@ -62,6 +62,10 @@ public:
     //!< \param unit The destroyed unit.
     virtual void OnUnitDestroyed(const Unit*) {}
 
+    //!  Called when a neutral unit is created. For example, mineral fields observed for the first time
+    //!< \param unit The observed unit.
+    virtual void OnNeutralUnitCreated(const Unit*) {}
+
     //! Called when a Unit has been created by the player.
     //!< \param unit The created unit.
     virtual void OnUnitCreated(const Unit*) {}
@@ -77,9 +81,14 @@ public:
     //!< \param upgrade The completed upgrade.
     virtual void OnUpgradeCompleted(UpgradeID) {}
 
-    //! Called when the unit in the previous step had a build progress less than 1.0 but is greater than or equal to 1.0 in the current step.
+    //! Called when the unit in the previous step had a build progress less than 1.0 but is greater than or equal to 1.0 in
+    // !the current step.
     //!< \param unit The constructed unit.
     virtual void OnBuildingConstructionComplete(const Unit*) {}
+
+    //! Called when the unit in the current observation has lower health or shields than in the previous observation.
+    //!< \param unit The damaged unit.
+    virtual void OnUnitDamaged(const Unit*) {}
 
     //! Called when a nydus is placed.
     virtual void OnNydusDetected() {}
--- a/include/sc2api/sc2_unit.h
+++ b/include/sc2api/sc2_unit.h
@@ -9,6 +9,7 @@
 #include "sc2_typeenums.h"
 #include <vector>
 #include <unordered_map>
+#include <unordered_set>
 #include <string>
 #include <stdint.h>
 
@@ -214,6 +215,17 @@ public:
     void ClearExisting();
     bool UnitExists(Tag tag);
 
+    const Units& GetNewUnits() const noexcept { return units_newly_created_; };
+    const Units& GetUnitsEnteringVision() const noexcept { return units_entering_vision_; };
+    const Units& GetCompletedBuildings() const noexcept { return units_constructed_; };
+    const Units& GetDamagedUnits() const noexcept { return units_damaged_; };
+    const std::unordered_set<const Unit*>& GetIdledUnits() const noexcept { return units_idled_; };
+
+    void AddUnitEnteredVision(const Unit *u) { units_entering_vision_.push_back(u); }
+    void AddCompletedBuilding(const Unit *u) { units_constructed_.push_back(u); }
+    void AddUnitIdled(const Unit *u) { if (u->alliance == Unit::Alliance::Self) units_idled_.insert(u); }
+    void AddUnitDamaged(const Unit *u) { units_damaged_.push_back(u); }
+
 private:
     void IncrementIndex();
 
@@ -222,8 +234,13 @@ private:
     // std::array<Unit, ENTRY_SIZE>
     std::vector<std::vector<Unit> > unit_pool_;
     PoolIndex available_index_;
-    std::unordered_map<Tag, Unit*> tag_to_unit_;
-    std::unordered_map<Tag, Unit*> tag_to_existing_unit_;
+    std::unordered_map<Tag, Unit *> tag_to_unit_;
+    std::unordered_map<Tag, Unit *> tag_to_existing_unit_;
+    Units units_newly_created_;
+    Units units_entering_vision_;
+    Units units_constructed_;
+    Units units_damaged_;
+    std::unordered_set<const Unit*> units_idled_;
 };
 
-}
+        }
--- a/src/sc2api/sc2_client.cc
+++ b/src/sc2api/sc2_client.cc
@@ -86,7 +86,6 @@ public:
 
     // Game state info.
     UnitPool unit_pool_;
-    std::unordered_map<Tag, Unit> units_previous_map_;
     uint32_t current_game_loop_;
     uint32_t previous_game_loop;
     RawActions raw_actions_;
@@ -578,15 +577,9 @@ bool ObservationImp::UpdateObservation() {
     if (observation_raw.HasErrors()) {
         return false;
     }
-
-    units_previous_map_.clear();
-    unit_pool_.ForEachExistingUnit([&](Unit& unit) {
-        units_previous_map_[unit.tag] = unit;
-    });
-
+    
     unit_pool_.ClearExisting();
-
-    Convert(observation_raw, unit_pool_, current_game_loop_);
+    Convert(observation_raw, unit_pool_, current_game_loop_, previous_game_loop);
 
     // Remap ability ids in orders.
     unit_pool_.ForEachExistingUnit([&](Unit& unit) {
@@ -1408,16 +1401,18 @@ public:
     bool PollResponse() override;
     bool ConsumeResponse() override;
 
-    bool IssueEvents(const std::vector<Tag>& commands = {}) override;
     void OnGameStart() override;
 
     void Error(ClientError error, const std::vector<std::string>& errors = {}) override;
     void ErrorIf(bool condition, ClientError error, const std::vector<std::string>& errors = {}) override;
 
+    bool IssueEvents(const std::vector<Tag>& commands = {}) override;
     void IssueUnitDestroyedEvents();
     void IssueUnitAddedEvents();
-    void IssueIdleEvent(const Unit* unit, const std::vector<Tag>& commands);
-    void IssueBuildingCompletedEvent(const Unit* unit);
+    void IssueIdleEvents(const std::vector<Tag>& commands);
+    void IssueBuildingCompletedEvents();
+    void IssueUnitDamagedEvents();
+
     void IssueAlertEvents();
     void IssueUpgradeEvents();
 
@@ -2036,74 +2031,62 @@ void ControlImp::IssueUnitDestroyedEvents() {
 }
 
 void ControlImp::IssueUnitAddedEvents() {
-    observation_imp_->unit_pool_.ForEachExistingUnit([&](sc2::Unit& unit) {
-        auto found = observation_imp_->units_previous_map_.find(unit.tag);
-        if (found != observation_imp_->units_previous_map_.end()) {
-            return;
-        }
 
-        if (unit.alliance == Unit::Alliance::Enemy && unit.display_type == Unit::DisplayType::Visible) {
-            client_.OnUnitEnterVision(&unit);
+    for (auto unit : observation_imp_->unit_pool_.GetNewUnits()) {
+        if (unit->alliance == Unit::Alliance::Self) {
+            client_.OnUnitCreated(unit);
         }
-        else if (unit.alliance == Unit::Alliance::Self) {
-            client_.OnUnitCreated(&unit);
+        else if (unit->alliance == Unit::Alliance::Neutral && unit->display_type == Unit::DisplayType::Visible) {
+            client_.OnNeutralUnitCreated(unit);
         }
-    });
+    }
+
+    for (auto unit : observation_imp_->unit_pool_.GetUnitsEnteringVision())
+        if (unit->alliance == Unit::Alliance::Enemy && unit->display_type == Unit::DisplayType::Visible) {
+            client_.OnUnitEnterVision(unit);
+        }
 }
 
-void ControlImp::IssueIdleEvent(const Unit* unit, const std::vector<Tag>& commands) {
-    if (!unit || !unit->orders.empty() || unit->build_progress < 1.0f) {
-        return;
+void ControlImp::IssueUnitDamagedEvents() {
+    for (auto const *u : observation_imp_->unit_pool_.GetDamagedUnits()) {
+        client_.OnUnitDamaged(u);
     }
-    // Lookup unit from previous map.
-    auto found = observation_imp_->units_previous_map_.find(unit->tag);
+}
 
-    // If it's not in the previous map it's a new unit with new orders so trigger the OnIdle event.
-    if (found == observation_imp_->units_previous_map_.end()) {
-        client_.OnUnitIdle(unit);
-        return;
-    }
+void ControlImp::IssueIdleEvents(const std::vector<Tag>& commands) {
 
-    // Otherwise get that unit from the previous list and verify it's state changed to idle.
-    const Unit& unit_previous = found->second;
-
-    if (!unit_previous.orders.empty()) {
-        client_.OnUnitIdle(unit);
-        return;
-    }
-
-    // If the unit had less than 1.0 build progress in the last stop this is the first time it's active.
-    if (unit_previous.build_progress < 1.0f) {
-        client_.OnUnitIdle(unit);
-        return;
-    }
-
-    // Iterate the issued commands, if a unit exists in that list but does not currently have orders
-    // the order must have failed. Reissue the OnUnitIdle event in that case.
+    auto& unit_pool = observation_imp_->unit_pool_;
+    // identify idled units where commands were issued last step, but units have no orders now (maybe failed, maybe executed instantly)
     for (auto t : commands) {
-        if (t == unit->tag) {
-            client_.OnUnitIdle(unit);
-            break;
-        }
+        const auto* unit = unit_pool.GetExistingUnit(t);
+        if (unit && unit->orders.empty())
+            unit_pool.AddUnitIdled(unit);
     }
+
+    // add newly created units (if they are completed)
+    for (auto const* u : unit_pool.GetNewUnits()) 
+      if (u->build_progress >= 1.0f && u->orders.empty()) {
+        unit_pool.AddUnitIdled(u);
+      }
+
+    // add completed buildings
+    for (auto const* u : unit_pool.GetCompletedBuildings()) 
+      if (u->build_progress >= 1.0f) {
+        unit_pool.AddUnitIdled(u);
+      }
+
+    // send only one idle event for any unit in any frame
+    for (auto const* u : unit_pool.GetIdledUnits()) {
+      client_.OnUnitIdle(u);
+    }
+
 }
 
-void ControlImp::IssueBuildingCompletedEvent(const Unit* unit) {
-    // If the units build progress is complete but it previously wasn't call construction complete
-    if (!unit || unit->build_progress < 1.0f) {
-        return;
-    }
-
-   auto found = observation_imp_->units_previous_map_.find(unit->tag);
-
-    if (found == observation_imp_->units_previous_map_.end()) {
-        return;
-    }
-
-    const Unit& unit_previous = found->second;
-    if (unit_previous.build_progress < 1.0f) {
-        client_.OnBuildingConstructionComplete(unit);
-    }
+void ControlImp::IssueBuildingCompletedEvents() {
+    for (auto unit : observation_imp_->unit_pool_.GetCompletedBuildings())
+        if (unit->alliance == Unit::Alliance::Self) {
+            client_.OnBuildingConstructionComplete(unit);
+        }
 }
 
 void ControlImp::IssueAlertEvents() {
@@ -2145,15 +2128,11 @@ bool ControlImp::IssueEvents(const std::vector<Tag>& commands) {
 
     IssueUnitDestroyedEvents();
     IssueUnitAddedEvents();
-
-    Units units = observation_imp_->GetUnits(Unit::Alliance::Self);
-    for (const auto& unit : units) {
-        IssueIdleEvent(unit, commands);
-        IssueBuildingCompletedEvent(unit);
-    }
-
+    IssueBuildingCompletedEvents();
+    IssueIdleEvents(commands);
     IssueUpgradeEvents();
     IssueAlertEvents();
+    IssueUnitDamagedEvents();
 
     // Run the users OnStep function after events have been issued.
     client_.OnStep();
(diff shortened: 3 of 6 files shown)


include/sc2api/sc2_client.h
```diff
@@ -62,6 +62,10 @@ class ClientEvents {
     //!< \param unit The destroyed unit.
     virtual void OnUnitDestroyed(const Unit*) {}
 
+    //!  Called when a neutral unit is created. For example, mineral fields observed for the first time
+    //!< \param unit The observed unit.
+    virtual void OnNeutralUnitCreated(const Unit*) {}
```
alkurbatov: To be honest I am not sure whether this event is needed. It could be useful, but requires some time to think about.

include/sc2api/sc2_unit.h
```diff
@@ -214,6 +215,17 @@ class UnitPool {
     void ClearExisting();
     bool UnitExists(Tag tag);
 
+    inline const std::vector<Unit const*>& GetNewUnits() const noexcept { return units_newly_created_; };
```
alkurbatov: Style:
`inline` is not ended here as the methods defined in the *.h field are lined by default, if I not mistaken.

include/sc2api/sc2_unit.h
```diff
@@ -214,6 +215,17 @@ class UnitPool {
...
+    inline const std::vector<Unit const*>& GetUnitsEnteringVision() const noexcept { return units_entering_vision_; };
```
alkurbatov: Style: please, use `const Unit*` instead of `Unit const*`.

src/sc2api/sc2_client.cc
```diff
@@ -86,7 +86,7 @@ class ObservationImp : public ObservationInterface {
 
     // Game state info.
     UnitPool unit_pool_;
-    std::unordered_map<Tag, Unit> units_previous_map_;
+    std::unordered_map<Tag, Unit> units_previous_map_; // this is now unused
```
alkurbatov: If it is unused, can we remove it?

src/sc2api/sc2_client.cc
```diff
@@ -521,8 +521,8 @@ bool ObservationImp::UpdateObservation() {
         return false;
     }
 
-    uint32_t next_game_loop = observation_->game_loop();
-    bool is_new_frame = next_game_loop != current_game_loop_;
+    const uint32_t next_game_loop = observation_->game_loop();
```
alkurbatov: Style: no need in marking `bool` and `int` as const, those are copies so changes in it don't matter.
Here and in other places.

src/sc2api/sc2_client.cc
```diff
@@ -578,15 +578,17 @@ bool ObservationImp::UpdateObservation() {
     if (observation_raw.HasErrors()) {
         return false;
     }
+
```
alkurbatov: Style: Extra empty line.

src/sc2api/sc2_client.cc
```diff
@@ -2035,73 +2039,60 @@ void ControlImp::IssueUnitDestroyedEvents() {
...
-    // Lookup unit from previous map.
-    auto found = observation_imp_->units_previous_map_.find(unit->tag);
 
-    // If it's not in the previous map it's a new unit with new orders so trigger the OnIdle event.
-    if (found == observation_imp_->units_previous_map_.end()) {
-        client_.OnUnitIdle(unit);
-        return;
+    for (auto unit : observation_imp_->unit_pool_.GetUnitsEnteringVision())
+    {
+        if (unit->alliance == Unit::Alliance::Enemy && unit->display_type == Unit::DisplayType::Visible)
```
alkurbatov: What about the shadows of the cloaked units?

src/sc2api/sc2_client.cc
```diff
@@ -2035,73 +2039,60 @@ void ControlImp::IssueUnitDestroyedEvents() {
...
+    // add newly created units (if they are completed)
+    for (auto const* u : unit_pool.GetNewUnits())
+        if (u->build_progress == 1.0f && u->orders.empty())
+            unit_pool.AddUnitIdled(u);
 
-   auto found = observation_imp_->units_previous_map_.find(unit->tag);
+    // add completed buildings (check for orders, to avoid possibility of classifying a newly constructed stargate that is
+    // transforming to warpgate as idle -- not sure that happens though)
+    for (auto const* u : unit_pool.GetCompletedBuildings())
+        if (u->build_progress == 1.0f && u->orders.empty())
```
alkurbatov: `u->build_progress == 1.0f` could not work properly for the Terran add ons. I heard about such issues case `Suvorov-bot` had similar logic. It was said that the progress is `>1.0f` in case of add ons. Please, check this case in advance.

src/sc2api/sc2_client.cc
```diff
@@ -2035,73 +2039,60 @@ void ControlImp::IssueUnitDestroyedEvents() {
...
     }
-}
 
-void ControlImp::IssueBuildingCompletedEvent(const Unit* unit) {
-    // If the units build progress is complete but it previously wasn't call construction complete
-    if (!unit || unit->build_progress < 1.0f) {
-        return;
-    }
+    // add newly created units (if they are completed)
+    for (auto const* u : unit_pool.GetNewUnits())
```
alkurbatov: Style: please add `{}` to the cycle body. Here and in other places.

src/sc2api/sc2_client.cc
```diff
@@ -2035,73 +2039,60 @@ void ControlImp::IssueUnitDestroyedEvents() {
...
+    for (auto const* u : unit_pool.GetCompletedBuildings())
+        if (u->build_progress == 1.0f && u->orders.empty())
+            unit_pool.AddUnitIdled(u);
 
-    if (found == observation_imp_->units_previous_map_.end()) {
-        return;
-    }
+    // send only one idle event for any unit in any frame
+    auto const& idled_units = unit_pool.GetIdledUnits();
+    std::for_each(idled_units.begin(), idled_units.end(), [this](auto const * u) {client_.OnUnitIdle(u); });
```
alkurbatov: Style: this is a bit inconsistent now as the code above uses simple `for` and here we use `for_each`.

src/sc2api/sc2_proto_to_pods.cc
```diff
@@ -202,9 +207,15 @@ bool Convert(const ObservationRawPtr& observation_raw, UnitPool& unit_pool, uint
         unit->is_on_screen = observation_unit.is_on_screen();
         unit->is_blip = observation_unit.is_blip();
 
-        unit->health = observation_unit.health();
+        const auto health = observation_unit.health();
+        if (health < unit->health)
+            unit_pool.AddUnitDamaged(unit);
```
alkurbatov: What about units which are under construction and under attack simultaneously? Maybe there is a trick to handle such cases too?

src/sc2api/sc2_proto_to_pods.cc
```diff
@@ -202,9 +207,15 @@ bool Convert(const ObservationRawPtr& observation_raw, UnitPool& unit_pool, uint
...
+        unit->health = health;
         unit->health_max = observation_unit.health_max();
         unit->shield = observation_unit.shield();
+        const auto shield = observation_unit.shield();
+        if (shield < unit->shield)
+            unit_pool.AddUnitDamaged(unit);
```
alkurbatov: Another question here: how often such events are issued?

alkurbatov:
Sorry, was too busy with my job for several weeks and thank you for the PR!

include/sc2api/sc2_client.h
mboedigh: My bot has a complete map of all of the neutral "snapshot" units at the start of the game. These are used for planning purposes. Bases, for example, have a vector of mineral resources they control. The trouble is, that I can't send a worker to a placeholder, because 1) it isn't possible to interact with it and 2) by the time the worker gets there the unit.tag and (unit *) have changed, so it is in fact a different unit. I use the single notification that the unit is encountered to replace the placeholder unit with an actual unit. Without the notification, it is challenging (older versions of my code had implemented it)
Final note: There are notifications for every other unit: OnUnitEnterVision for enemy and OnUnitCreated for friendly, 
Let me know your thoughts

include/sc2api/sc2_unit.h
mboedigh:
```suggestion
    const std::vector<Unit const*>& GetNewUnits() const noexcept { return units_newly_created_; };
```

mboedigh: I agree with all these comments and style comments as well.. Should I do them here with suggestions? or by resubmitting a PR?

src/sc2api/sc2_client.cc
mboedigh: Unless you see a reason, it can be deleted. There is another reference to it that will also need to be deleted. I can do that after this PR is resolved

src/sc2api/sc2_proto_to_pods.cc
mboedigh: good point. I don't know how the game handles health during construction. If you don't know I'll do some testing. If the current health of an un-attacked building is proportional to progress it would suggest possible approaches.

src/sc2api/sc2_proto_to_pods.cc
mboedigh: once for every observation in which a unit takes damage. It is often. I have a batch version of this that sends out the vector of all friendly damaged units. for my own bot, I would rather have a map for is_damaged[unit *]. I think of using the notification to create such a map. But I'm not sure in the end, if that is the best approach. The whole notification system might be better implemented as a pull instead of a push. Maybe with a multi-index (ala boost) that allows users to setup indices that are important.

include/sc2api/sc2_client.h
alkurbatov: Ok, thank you for the detailed explanation. Let's give it a go.

include/sc2api/sc2_unit.h
alkurbatov: You should do combination of `git rebase -I` (to add changes) + `git push --force` (to replace the old commits with the changed version). There is no need in recreating the PR.

src/sc2api/sc2_client.cc
alkurbatov: Ah, ok. I'll check it.

src/sc2api/sc2_proto_to_pods.cc
alkurbatov: I don't want to bring in boost as it usually causes additional difficulties for less experienced programmers. While it is relatively easy to install proper boost version on Linux and OS X, Windows is usually pain. So, unless we have very strong necessity, better to avoid usage of boost in this project.

src/sc2api/sc2_client.cc
mboedigh: This replicates existing behavior for OnUnitEnterVision(). The display_type is different than cloak. Let me know if you still see an issue.

src/sc2api/sc2_client.cc
mboedigh: I'll change to >= 1.0f just in case, as it can do no harm, but this replicates existing behavior.

src/sc2api/sc2_client.cc
mboedigh: your right. I think because my other lambdas weren't compiling for me. I'll use the simple for block for consistency.

src/sc2api/sc2_proto_to_pods.cc
mboedigh: yeah, maybe best not to complicate things.

include/sc2api/sc2_unit.h
mboedigh: I'm not familar enough PRs and forks know what this actually does. My repository is a fork of yours, and I created the PR from my remote repository on github. If I do the above git commands from from my local repository I assume it will only alter my own fork on github, right?

alkurbatov: >If I do the above git commands from from my local repository I assume it will only alter my own fork on github, right?

Yes, that's why it is safe.

src/sc2api/sc2_client.cc
alkurbatov: There is nothing bad in extending the current behaviour, but I suggest to check this assumption first. I can replicate it in my bot, but I need to teach it how to build add ons. Will take some time.

src/sc2api/sc2_client.cc
alkurbatov: I personally prefer `for_each` although it could be harder to implement and read. So, it would be cool to switch to `for_each` whenever possible.

src/sc2api/sc2_client.cc
mboedigh: I checked and the orders vector is empty when the gateway is complete, so it gets an OnIdle event even if it is not truly idle (i.e. it will morph into a warpgate automatically).
So I removed the check and comment

src/sc2api/sc2_client.cc
mboedigh: I agree it is harder to read and write. Also, for_each is stylistically inconsistent with the current sc2api (it is not used anywhere). Maybe we can wait on this?

src/sc2api/sc2_client.cc
alkurbatov: Yes, I see it in my bot too. Pretty inconvenient.

src/sc2api/sc2_client.cc
alkurbatov: Ok, no problem.

include/sc2api/sc2_unit.h
mboedigh: scary stuff ;) I pushed changes to the branch on github that the PR was created from. Does this mean you can see them automatically now? 
That should be all the changes from all your comments for this PR. If I missed anything let me know.

src/sc2api/sc2_client.cc
alkurbatov: >u->build_progress == 1.0f could not work properly for the Terran add ons. 

At least I cannot reproduce it. Checked on barracks reactor and techlab.

mboedigh: build_progress == 1.0f seems to work fine for me too. Checking for >= 1.0f is bullet proof, and guards against future changes to client. Currently the client side must take care of setting build_progress to at most 1.0f. I believe the current branch is >=, but if you prefer ==, let me know (or you can change yourself)

src/sc2api/sc2_client.cc
mboedigh: I deleted this in the branch on my fork. I (attempted) to rebase to the same commit used in the PR. Let me know if you cannot see it and, if not, what to do to fix it

src/sc2api/sc2_client.cc
mboedigh: I believe these are done

src/sc2api/sc2_client.cc
mboedigh: Oops, I've been resolving some of these change requests, but maybe that is for you to do.

src/sc2api/sc2_client.cc
mboedigh: done

src/sc2api/sc2_proto_to_pods.cc
mboedigh: I looked into this. It is a linear relationship, but it is weird. The first frame after a structure is created it has 10% max_health plus a small delta. The delta is different that the delta that is added to each subsequent frame. I agree we can skip for now, with a little more effort the exact relationship should be decipherable

include/sc2api/sc2_unit.h
```diff
@@ -214,6 +215,17 @@ class UnitPool {
     void ClearExisting();
     bool UnitExists(Tag tag);
 
+    const std::vector<const Unit*>& GetNewUnits() const noexcept { return units_newly_created_; };
```
alkurbatov: There is a nice shortcut for the returned values here:
`17:include/sc2api/sc2_unit.h:202:typedef std::vector<const Unit*> Units`

Could make the things shorter.

include/sc2api/sc2_unit.h
```diff
@@ -222,8 +234,14 @@ class UnitPool {
     // std::array<Unit, ENTRY_SIZE>
     std::vector<std::vector<Unit> > unit_pool_;
     PoolIndex available_index_;
-    std::unordered_map<Tag, Unit*> tag_to_unit_;
-    std::unordered_map<Tag, Unit*> tag_to_existing_unit_;
+    std::unordered_map<Tag, Unit *> tag_to_unit_;
```
alkurbatov: Extra space. Here and below.

include/sc2api/sc2_unit.h
```diff
@@ -9,6 +9,7 @@
 #include "sc2_typeenums.h"
 #include <vector>
 #include <unordered_map>
```
alkurbatov: Do we still need `#include <unordered_map>` after removal of `units_previous_map_`?

src/sc2api/sc2_client.cc
```diff
@@ -2035,73 +2030,69 @@ void ControlImp::IssueUnitDestroyedEvents() {
...
 void ControlImp::IssueUnitAddedEvents() {
-    observation_imp_->unit_pool_.ForEachExistingUnit([&](sc2::Unit& unit) {
-        auto found = observation_imp_->units_previous_map_.find(unit.tag);
-        if (found != observation_imp_->units_previous_map_.end()) {
-            return;
-        }
 
-        if (unit.alliance == Unit::Alliance::Enemy && unit.display_type == Unit::DisplayType::Visible) {
-            client_.OnUnitEnterVision(&unit);
+    for (auto unit : observation_imp_->unit_pool_.GetNewUnits())
```
alkurbatov: Style:
Seems that the preferred way it to do:
```cpp
for (auto unit : observation_imp_->unit_pool_.GetNewUnits()) {
```

It is ok to leave it as is.

src/sc2api/sc2_client.cc
```diff
@@ -2035,73 +2030,69 @@ void ControlImp::IssueUnitDestroyedEvents() {
...
-    auto found = observation_imp_->units_previous_map_.find(unit->tag);
 
-    // If it's not in the previous map it's a new unit with new orders so trigger the OnIdle event.
-    if (found == observation_imp_->units_previous_map_.end()) {
-        client_.OnUnitIdle(unit);
-        return;
+    for (auto unit : observation_imp_->unit_pool_.GetUnitsEnteringVision())
+    {
+        if (unit->alliance == Unit::Alliance::Enemy && unit->display_type == Unit::DisplayType::Visible) {
+            client_.OnUnitEnterVision(unit);
```
alkurbatov: Style:
You can safely drop brackets here, i.e.
```cpp
if (unit->alliance == Unit::Alliance::Enemy && unit->display_type == Unit::DisplayType::Visible)
             client_.OnUnitEnterVision(unit);
```

But it is ok to leave it as is.

src/sc2api/sc2_client.cc
```diff
@@ -2035,73 +2030,69 @@ void ControlImp::IssueUnitDestroyedEvents() {
...
+      if (u->build_progress >= 1.0f) {
+        unit_pool.AddUnitIdled(u);
+      }
+    }
 
-    if (found == observation_imp_->units_previous_map_.end()) {
-        return;
+    // send only one idle event for any unit in any frame
+    auto const& idled_units = unit_pool.GetIdledUnits();
+    for (auto const* u : idled_units) {
```
alkurbatov: Style:
Should be 
```cpp
     for (auto const* u : unit_pool.GetIdledUnits()) {
```
like several line earlier.

src/sc2api/sc2_proto_to_pods.cc
```diff
@@ -185,7 +185,12 @@ bool Convert(const ObservationRawPtr& observation_raw, UnitPool& unit_pool, uint
         unit->pos.z = pt.z();
         unit->facing = observation_unit.facing();
         unit->radius = observation_unit.radius();
-        unit->build_progress = observation_unit.build_progress();
+
+        const auto bp = observation_unit.build_progress();
+        if (bp == 1.0f && unit->build_progress < 1.0f)
```
alkurbatov: `bp >= 1.0f`?

mboedigh:
NP and thanks.
I’m traveling and will make the suggested changes when I get back at the end of the month.

include/sc2api/sc2_unit.h
mboedigh: yes, it is used elsewhere.

mboedigh:
To clarify. Should I make a new commit with one message (e.g. follows), and then rebase so there is one commit for all it?  I want to be sure, because I’m don’t know if I can undo a rebase.

feat: On Neutral Unit Created, #33
OnNeutralUnitCreated is now called when a neutral unit, such as a mineral field, is observed for the first time. Prior to that time, the mineral field is a snapshot, and other units cannot interact with it. This callback provides a mechanism to know that the snapshot is replaced with a visible unit that other units can interacted with. It is analogous to OnUnitEnterVision and OnUnitCreated

feat: On Unit Damaged
OnUnitDamaged is called whenever a friendly unit has fewer health or shield points relative to the previous observation.

perf: UnitPool now caches units encountering certain events
Previously sc2 would iterate through every unit, look it up in a map, and then determine whether it meets criteria to trigger an event each event and each observation. Now if a units meets the criteria it is stored in a vector when it is first retrieved, avoiding the iterations and mapping altogether.

fix: hang due to OnUnitIdle event being called multiple times for the same unit
OnUnitIdle will only be called once for any given unit.

alkurbatov:
Ideally, there should be several commits with proper titles, but you can create one commit if it is simpler.

mboedigh:
I pushed changes to my fork that was used for this PR. I tried to "squash" extraneous commits, but am totally unsure whether I got all of them.  Please let me know.

alkurbatov:
It seems that everything is broken now, see the CI results.

Regarding the commits: you shouldn't send the merge commits and intermediate commits as it is details of intermediate work.

mboedigh:
It was working on my side and I think I pushed wrong. I say that because I had to resolve conflicts (and resolved them incorrectly). I believe I should have used git push –force to replace my previous branch.

I ‘squashed’ all commits except one and it compiles and passes on my side. However, the commit appears like it is coming from a node tagged with
HEAD, dev, master and v1.1.0.

This was extremely painful and a lot of reworking, but I think I have a better understanding of git in general, and rebase in particular. I hope it is all ok now

Let me know if anything is awry

alkurbatov:
@mboedigh seems that I can't merge the PR due to
>Conflicting files
>include/sc2api/sc2_unit.h

mboedigh:
I can't explain that. The branch my fork calls _dev_ has the UnitFilters (e.g. IsUnit) in sc2_unit.h and yours has it in a separate file. Apparently the branch my fork calls dev and the upstream calls dev are divergent. I don't know how that even happens. Or more importantly, how to fix it. I can try changing sc2_unit.h so that it matches yours and push that (after rebasing), but I won't be able to test it because it doesn't compile on my side. If merge is successful I will try to delete my fork and start over with the your upstream fork

mboedigh:
or I could try to fetch your current dev branch and try to rebase onto that.

alkurbatov:
> The branch my fork calls dev has the UnitFilters (e.g. IsUnit) in sc2_unit.h and yours has it in a separate file.

Thats because I moved the filters in a separate files. 
I think I can apply you PR on the latest dev branch manually.

mboedigh:
I've attempted to apply my patch to the head of your current dev branch. I believe it compiles and tests ok and doesn't have any extra commits. The CI seems hung or is taking a long time to complete.

alkurbatov:
Hmm, CI is definitely doesn't fire, might be a bug somewhere. Ok, I'll take a close look tomorrow and merge the request without CI. We can always catch broken build later.

mboedigh:
I already found a bug. Unit.last_seen_game_loop was not properly initialized during Unit creation (regression). I can try to push that and see if it triggers a new CI. Let me know

alkurbatov:
Yes, please, do.

alkurbatov:
Checked the compilation manually on OS X. Looks good to me.
The rest will be checked by Travis after merge.

@mboedigh many thanks for patience and the great PR!

PR #38, opened 2019-10-08
New setting allowing the use of raw_affects_selection

RaphaelRoyerRivard:
feat: New coordinator setting allowing users to choose to reduce the number of generated actions


alkurbatov:
Thank you! Looks good to me.

Issue #34, opened 2019-09-26
define general_holdposition enum


mboedigh:
This may be a duplicate of fix holdposition, but I couldn't tell based on the description. 
Like MOVE, PATROL, HOLDPOSITION has a new enum that appears to be the generalization of specialized commands (If I understand correctly, which is questionable here). 
It looks like move and patrol enum values were _changed_, rather than adding new enums for the generalized values. I propose doing the same for HOLDPOSITION (change value from 18 to 3793)

alkurbatov:
Yep, this is mostly the dup of 'fix holdposition'. 

A small portion of history :)
I thought about adding 'general' versions of the same commands as it allows to support older game versions. But so far it is unlikely that we can (should) support something older than 4.10.
From the other point of view, having both 'general' and 'basic' versions of the same command could be quite confusing for a developer.

mboedigh:
so, if we want to learn from an old replay, will the commands be different?

alkurbatov:
Actually, a good question. Yes, might be different if the replay game version < 4.10. 
It is a good idea to check the replays.

mboedigh:
:) I can't get my replays working (issue #25), which is why I'm working on my scripted bot. If you can point me in the right direction for fixing 25 I would be glad to help. I'm just totally stumped by it.

alkurbatov:
Will be fixed together with #43 in the next release.

alkurbatov:
Fixed.

PR #41, opened 2019-10-17
Added and updated some IDs

RaphaelRoyerRivard:
I made a small mistake and reverted a commit, then recreated it. Will it cause problem?


alkurbatov:
Yes, please, get rid of the merge commits. The history should be clear, otherwise it could be difficult to find out the reason of changes.

include/sc2api/sc2_typeenums.h
```diff
@@ -456,6 +456,7 @@ namespace sc2 {
         EFFECT_VOIDRAYPRISMATICALIGNMENT = 2393,  // Target: None.
         EFFECT_WIDOWMINEATTACK = 2099,  // Target: Unit.
         EFFECT_YAMATOGUN = 401,   // Target: Unit.
+        GENERAL_MOVE = 3794,  // Target: Point.
```
alkurbatov: If you define GENERAL_MOVE it is ok to keep it as 3794, but you should use the old value for the MOVE command.

RaphaelRoyerRivard:
I do not use that command, it's just that when debugging, I saw that
issuing a MOVE command sets an order to the unit with that GENERAL_MOVE
ability ID. Since it was not in the list, I had to search to find what that
ID meant. I thought it might be useful for someone else also.

alkurbatov:
Here is a longer discussion which also gives a possible reason why we should keep old ID for the old command: #34

@mboedigh FYI

RaphaelRoyerRivard:
I didn't remove the old MOVE command.

alkurbatov:
I know, I mean: if you introduce GENERAL_MOVE, you should change id of the original MOVE command.

RaphaelRoyerRivard:
Since the old MOVE command still works, I don't understand why we would change it. For what value should I change it?

alkurbatov:
The old move command doesn't work in 4.10, but works in older version (with old ID). This is needed mostly for replay parsing and backward compatibility. See the discussion in the issue I linked above.
Feel free to drop this change, as I am going to add other generalized abilities.

include/sc2api/sc2_typeenums.h
```diff
@@ -866,6 +867,7 @@ namespace sc2 {
         CARRYHARVESTABLEVESPENEGEYSERGAS = 273,
         CARRYHARVESTABLEVESPENEGEYSERGASPROTOSS = 274,
         CARRYHARVESTABLEVESPENEGEYSERGASZERG = 275,
+        RAVENSCRAMBLERMISSILE = 277,
```
alkurbatov: Please, add the new ID to the `BuffIDToName` function (`src/sc2api/sc2_typeenums.cc`).

PR #47, opened 2019-12-23
Make == and != operators const functions for Point2D and Point3D

mboedigh:
fix: == and != operators in Point2D and Point3D should be const functions to implement equal_to interface

### DIFF
--- a/include/sc2api/sc2_common.h
+++ b/include/sc2api/sc2_common.h
@@ -33,8 +33,8 @@ struct Point3D {
     Point3D& operator*=(float rhs);
     Point3D& operator/=(float rhs);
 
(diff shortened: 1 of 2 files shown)


mboedigh:
sorry, this PR is apparently trying to merge into master instead of dev, but they both seem to be at the same point. Is that a problem?

alkurbatov:
> but they both seem to be at the same point.

Because I synced it.

Usually I develop in the dev branch (an attempt to make the master branch more stable), but we can merge this PR to master as it is harmless.

alkurbatov:
I think I found the reason of hanging CI. Maybe it'll help in the future.
Thank you for PR! :)

PR #50, opened 2020-01-27
Fix get food workers

mboedigh:
fix: GetFoodWorkers() should return uint32_t

### DIFF
--- a/include/sc2api/sc2_client.h
+++ b/include/sc2api/sc2_client.h
@@ -88,8 +88,8 @@ public:
 
     //! Called when the unit in the current observation has lower health or shields than in the previous observation.
     //!< \param unit The damaged unit.
-    virtual void OnUnitDamaged(const Unit*, float) {}
-     
+    virtual void OnUnitDamaged(const Unit*) {}
+
     //! Called when a nydus is placed.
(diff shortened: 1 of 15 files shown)


alkurbatov:
Looks like we have two unnecessary commits here.

mboedigh:
This is weird. 
1) You already merged Point_Const_EqualTo. it doesn't have any new changes and is in your master branch (ccbb74f246df472df5e705b318b2e17f25573ff6 is the merge pull request commit). I think that was around the time CI went belly up. 
2) While I have a branch to update OnUnitDamaged, I was going to propose that separately. That commit is NOT in my logs on origin.
image

Now what do I do? On my local, FixGetFoodWorkers comes after OnUnitDamaged, so I'll try to rebase OnUnitDamaged to come after and then push.

alkurbatov:
It could happen e.g. if you did rebase and touched those commits in some way (e.g. changed the hash).

>Now what do I do? On my local, FixGetFoodWorkers comes after OnUnitDamaged, so I'll try to rebase OnUnitDamaged to come after and then push.

I recommend to create a new branch based on cpp-sc2's dev branch and cherry-pick your commit. It should be enough.

PR #56, opened 2020-02-11
feat: Upgraded type enums for 4.11.3

Cryptyc:


include/sc2api/sc2_typeenums.h
```diff
@@ -368,7 +396,6 @@ namespace sc2 {
         CANCEL_ADEPTSHADEPHASESHIFT = 2596,  // Target: None.
         CANCEL_BARRACKSADDON = 451,   // Target: None.
         CANCEL_BUILDINPROGRESS = 314,   // Target: None.
-        CANCEL_CREEPTUMOR = 1763,  // Target: None.
```
alkurbatov: CANCEL_CREEPTUMOR and EFFECT_INFESTEDTERRANS were removed. Is it ok? I believe we should keep it fr older versions.

Cryptyc: We can if you like, it could cause some confusion on newer versions, but I think there might be a few around like that anyway.  
The other option is to separate the branches for compatibility versions, which might be a good option overall, as some of the changes might be problematic on older versions, but I'll go with whatever you think is best :)

RaphaelRoyerRivard:
I think having 2 branches separated on version compatibility would be wise
since authors like me would like to be able to have a version working on
4.10 for ai-arena and another one on 4.11 for sc2ai. I'd prefer not have to
insert conditions to check version when selecting a unit type id for
example.

include/sc2api/sc2_typeenums.h
alkurbatov: I would follow the branches approach, but it'll lead to problems with replay analysers as one could use older replays with the newest API. Because of it, I suggest to keep old IDs as is.

alkurbatov:
@RaphaelRoyerRivard in this case you'll have to check version somewhere in cmake and link with different precompiled libraries. The same problem imho.

RaphaelRoyerRivard:
Well at least this would done once and not at every place in the code where
it uses a type id that changed in the new version.

alkurbatov:
Need to think about the branches more, so far I am not sure mostly because of the replays parsing.

include/sc2api/sc2_typeenums.h
Cryptyc: cool, I'll update the PR tonight

Cryptyc:
Sounds good, will update tonight

PR #61, opened 2020-03-03
Include Damage in OnUnitDamaged events

mboedigh:
feat: Include the amount of health and shield damage in OnUnitDamaged events

### DIFF
--- a/src/sc2api/sc2_proto_to_pods.cc
+++ b/src/sc2api/sc2_proto_to_pods.cc
@@ -218,7 +218,7 @@ bool Convert(const ObservationRawPtr& observation_raw, UnitPool& unit_pool, uint
         unit->shield = cur_shield;
 
         if (damage > 0 || shield_damage > 0)
-            unit_pool.AddUnitDamaged(unit);
+            unit_pool.AddUnitDamaged(unit, damage, shield_damage);
 
         unit->health_max = observation_unit.health_max();
         unit->shield_max = observation_unit.shield_max();
--- a/include/sc2api/sc2_client.h
+++ b/include/sc2api/sc2_client.h
@@ -88,7 +88,9 @@ public:
 
     //! Called when the unit in the current observation has lower health or shields than in the previous observation.
     //!< \param unit The damaged unit.
-    virtual void OnUnitDamaged(const Unit*) {}
+    //!< \param health The change in health (damage is positive)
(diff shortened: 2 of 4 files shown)


src/sc2api/sc2_proto_to_pods.cc
```diff
@@ -218,7 +218,7 @@ bool Convert(const ObservationRawPtr& observation_raw, UnitPool& unit_pool, uint
         unit->shield = cur_shield;
 
         if (damage > 0 || shield_damage > 0)
-            unit_pool.AddUnitDamaged(unit);
+            unit_pool.AddUnitDamaged(unit, damage, shield_damage);
```
alkurbatov: Seems that `damage` or `shield_damage` could be negative. Shouldn't we report zeroes instead?

mboedigh: mathematically it is possible, but mechanically, I can't get it to behave that way. So far, it seems that if either type of damage is taken (net) the other does not heal.  specifically, if health decreases I haven't seen shields increase. (and, of course, the concern doesn't come in unless the unit has shields). 
In either case it depends on what we want to report. I see a couple choices: 
1) change the documentation to report change in health and shield whenever either is damaged. e.g. the health parameter is the change in health whenever health or shields decreases. (my preference).
2)  change the behavior to truncate any healing to 0. e.g. the health parameter is the damage to health whenever health or shields decreases. 
Let me know.

alkurbatov: I think the simplest solution is to use `std::max(value, 0)`.

mboedigh: Ok, To be clear, my question is about _how_ to report positive damage. That part is easy. My question is about whether it makes more sense to allow negative damage (if such a scenario exists). negative damage (i.e. healing) is real and valid information that might be useful. I'm not sure it makes more sense to remove this information for no particular reason, especially since it is (albeit trivially) less efficient to do so.  So my options were to change the documentation to be clear that negative damage might, in theory,  be reported OR change the code so that they are not possible. 
Can you confirm your choice?

alkurbatov: Unfortunately, I don't know. Lets change the documentation and keep the implementation as simple as possible. We can always tweak this behaviour according to API users feedback.

PR #63, opened 2020-05-04
fix: There was an issue with the IsTownHall filter that caused it to …

ddelamare:
…always return true when a unit was passed in.


alkurbatov:
@ddelamare thank you for the fix! I extended it a bit as we have additional unit filters in the dev branch as all developments and fixes happen there, see 8731e75f5c88190dcf6290b53520f3f2bb5dd3a9

Issue #18, opened 2019-09-13
Fix HOLDPOSITION command


alkurbatov:
- [ ] Fix the HOLDPOSITION command.
- [ ] Add a functional test.

alkurbatov:
Will be fixed together with https://github.com/alkurbatov/cpp-sc2/issues/43 in the next release.

alkurbatov:
Fixed.

PR #64, opened 2020-05-10
feat: Autogenerated ids

### LINKED ISSUE #43: Collect UNIT_TYPE_IDs automagically
Partly adopt approach of python-sc2 and build a tool which will correct the actual IDs for units, buffs etc.
Warning: We should preserve old ability IDs for backward compatibility (e. g. replay parsing).

alkurbatov:
Fixes #43, #34, #18.

Now we get ID of abilities autogenerated by https://github.com/cpp-sc2/codegen


alkurbatov:
Fixed: Use friendlyName to generate array of effects.

alkurbatov:
Fixed: Add prefixes to more units.

Issue #9, opened 2019-09-08
CONTRIBUTING


alkurbatov:
Hi, if you are reading this is because you pretty likely want to participate in the development of this API and curious about how to start.

You are in **cpp-sc2** core project, but actually, we have more repositories you can contribute to:
- **[codegen](https://github.com/cpp-sc2/codegen)** - type IDs codegeneration utility;
- **[commandcenter](https://github.com/cpp-sc2/commandcenter)** - original version of CommandCenter bot with integrated cpp-sc2;
- **[docker-sc2](https://github.com/cpp-sc2/docker-sc2)** - dockerized StarCraft II Linux client;
- **[scrubber](https://github.com/cpp-sc2/scrubber)** - scrubs type ids from a game client.
- **[blank-bot](https://github.com/cpp-sc2/blank-bot)** - starter bot for StarCraft II with integrated cpp-sc2.

## What can I do?

- [ ] Search for other issues tagged [with `good first issue` or `help wanted` tag](https://github.com/cpp-sc2/cpp-sc2/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22).
- [ ] Suggest new features.
- [ ] Vote for existing issues (issues with highest number of reactions are likely to be fixed first).
- [ ] Fix `// FIXME: ` comments in the code.
- [ ] Improve the documentation of the API.
- [ ] Add/Fix/Improve tutorials.
- [ ] Add/Fix/Improve new bot examples.
- [ ] Add/Fix/Improve tests.
- [ ] Help to test latest changes.
- [ ] Support this project on [Boosty](https://boosty.to/cpp-sc2) (preferred) or [Patreon](https://www.patreon.com/cppsc2). Support of the project allows to make releases more often and implement nice new features for developers.

If you have any suggestions, feel free to comment on this thread, or chat me over [Discord](https://discordapp.com/invite/Emm5Ztz) (the `#cpp` channel).

## What should I know before the start?

**Code guidelines**
We do our best to conform to [the Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) with the exception that we use four space tabs instead of two space tabs.

**Commit guidelines**
We follow the semantic versioning approach, this is important so your PR will appear on the Changelog 🎉 ! I guess you expect recognition and we will give it to you.
See [standard-version](https://github.com/conventional-changelog/standard-version#commit-message-convention-at-a-glance) for commit guidelines.

The list of supported commit message prefixes:

[...]

PR #99, opened 2021-03-23
Fix of zombie processes

procrash:
This patch encompasses several fixes to extract replay data concurrently in a larger manner.
When trying to do this using the CPP-SC2 API zombie process arose. 
There were also problems showing up with SC2 Headless clients that didn't respond anymore and the API just exited. 
To recognize this problem better an exception has been introduced.

It is now recommended to call waitForPID after the 
while (coordinator.Update());
loop has been finished.

As the waitForPID might wait forever it is advisory to introduce some kind of a watchdog facility with:

        std::cout << "Waiting for PID" << std::endl;

        std::thread watchDog = std::thread([&]() {
            std::this_thread::sleep_for(std::chrono::seconds(3));
            std::cout << "Terminating..." << std::endl;
            coordinator.terminate();
            std::cout << "Terminating done..." << std::endl;
        });

        coordinator.waitPID();
        watchDog.join();

### DIFF
--- a/examples/echo_actions.cc
+++ b/examples/echo_actions.cc
@@ -34,22 +34,32 @@ public:
         // For now, just show  the first action.
         const sc2::ActionRaw& action = actions[0];
 
-        // Show the index of the ability being used.
-        last_action_text_ = GetAbilityText(action.ability_id);
 
-        // Add targeting information.
-        switch (action.target_type) {
-            case sc2::ActionRaw::TargetUnitTag:
-                last_action_text_ += "\nTargeting Unit: " + std::to_string(action.target_tag);
-                break;
+        const sc2::ActionRawUnitCommand* ptrActionRawUnitCommand=dynamic_cast<const sc2::ActionRawUnitCommand*>(&action);
 
-            case sc2::ActionRaw::TargetPosition:
-                last_action_text_ += "\nTargeting Pos: " + std::to_string(action.target_point.x) + ", " + std::to_string(action.target_point.y);
-                break;
 
-            case sc2::ActionRaw::TargetNone:
-            default:
-                last_action_text_ += "\nTargeting self";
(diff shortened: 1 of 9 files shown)


alkurbatov:
Hi @procrash 
Thank you for the contribution!
Sorry for the late answer, life doesn't give time to catch breath.

I can't take this pR as is because of different reasons. Most of them described here: https://github.com/cpp-sc2/cpp-sc2/issues/9
However, I'll slowly check and merge the parts Which I think will be good addition to the API. Below in more details.

alkurbatov:
>Connection errors to SC2 client now throw an exception instead of asserting

Totally agree, asserts are not so informative. However, this commit has several style errors and the actual assert is still there. I'll fix the commit and merge it to the dev branch.

alkurbatov:
>Fixed RTTI information

Sorry can't agree with you on this.
In C++ there is no difference between `struct` and `class` except `public` and 'private' keywords used by default for inheritance and members declaration.
In other words the following two code blocks are equivalent:
```c++
struct A: B
{
  int c;
}
```
```c++
class A: public B
{
 public:
  int c;
}
```

Both are correct even, although I personally prefer the former.

Thus, I think I'll drop excessive `public` keywords, but keep the original approach.

alkurbatov:
>Added queue commands

Please, describe why do we need this change and which problem it solves?

alkurbatov:
>Added interface to terminate running coordinator process

Seems that it will not work on Windows which is stopper for this commit.
Also, I need additional information regarding it. Ideally, how-to reproduce the issue.

alkurbatov:
Closed as not ready to merge.

we will try to reimplement and adopt parts of this PR in https://github.com/cpp-sc2/cpp-sc2/issues/102 and https://github.com/cpp-sc2/cpp-sc2/issues/103

Issue #97, opened 2021-03-18
Add missing DebugIgnoreGas


ImpulseCloud:
**Is your feature request related to a particular part of the API (function, class, functor)? If so, specify it.**
Currently, DebugGameState enum has 'gas' in the DebugGameState Enum, that the API doesn't use.
'gas' is the only one in the Enum that was not given a Debug command to enable.
```
enum DebugGameState : int {
  show_map = 1,
  control_enemy = 2,
  food = 3,
  free = 4,
  all_resources = 5,
  god = 6,
  minerals = 7,
  gas = 8,
  cooldown = 9,
  tech_tree = 10,
  upgrade = 11,
  fast_build = 12
};
```
**Is your feature request related to a problem? Please describe.**
#85 is related. Wanted to ignore training costs (I'm assuming), but still include health/energy costs for abilities.

**Describe the solution you'd like**
The API could add a DebugIgnoreGas() which ignore gas costs, if you are trying to build armies without costs.

sc2_interfaces.h
```
//! Ignores gas costs.
virtual void DebugIgnoreGas() = 0;
```
sc2_client.cc
```
void DebugIgnoreGas() override;

void DebugImp::DebugIgnoreGas() {
    debug_state_.push_back(SC2APIProtocol::DebugGameState::gas);
}
```

alkurbatov:
Fixed.

PR #108, opened 2021-10-05
fix: missing DebugIgnoreGas

Nickrader:
I'm guessing the issue, Add missing DebugIgnoreGas #97 , was just a matter of adding in the code present in the conversation.


alkurbatov:
Thx!
Looks like it's all we need.

alkurbatov (approved):

PR #111, opened 2021-10-25
feat: Allow ActionImp::UnitCommand to use sc2::Tag

Nickrader:
I think I have it except for the abstraction.  In the tests I couldn't figure out a more concise way other than overriding the IssueUnitCommand() function, to explicitly call the Tag overloads of ActionImp::UnitCommand().  
I'll try and get my sticking points listed with the abstraction, soon.  Basically, when making the struct, if it is derived from ActionImp, then I can use the member function   GetRequestAction() to consolidate that first mess. 
```
    SC2APIProtocol::RequestAction* request_action = GetRequestAction();
    SC2APIProtocol::Action* action = request_action->add_actions();
    SC2APIProtocol::ActionRaw* action_raw = action->mutable_action_raw();
    SC2APIProtocol::ActionRawUnitCommand* tag_command = action_raw->mutable_unit_command();
```
But then I get a little lost in how the other parameters (abiliity, point, queued_command) get picked up.
Trying to default initialize the parameters in the struct constructor, seems to lead me into more overloads, which probably aren't needed. 
Now that I have the tests done and am understanding the inheritance in the tests, I'll take another few attempts at the abstraction and get my sticking points drawn up better.

### DIFF
--- a/include/sc2api/sc2_unit.h
+++ b/include/sc2api/sc2_unit.h
@@ -218,7 +218,10 @@ public:
 };
 
 typedef std::vector<const Unit*> Units;
+typedef std::vector<Tag> Tags;
 typedef std::unordered_map<Tag, size_t> UnitIdxMap;
+
+Tags ConvertToTags(const Units& units);
 
 struct UnitDamage {
     const Unit* unit;
--- a/src/sc2api/sc2_agent.cc
+++ b/src/sc2api/sc2_agent.cc
@@ -26,17 +26,23 @@ public:
     void UnitCommand(const Units& units, AbilityID ability, bool queued_command = false) override;
     void UnitCommand(const Units& units, AbilityID ability, const Point2D& point, bool queued_command = false) override;
     void UnitCommand(const Units& units, AbilityID ability, const Unit* target, bool queued_command = false) override;
+    void UnitCommand(Tag tag, AbilityID ability, bool queued_command = false) override;
+    void UnitCommand(Tag tag, AbilityID ability, const Point2D& point, bool queued_command = false) override;
+    void UnitCommand(Tag tag, AbilityID ability, const Tag target_tag, bool queued_command = false) override;
+    void UnitCommand(const Tags& tags, AbilityID ability, bool queued_command = false) override;
+    void UnitCommand(const Tags& tags, AbilityID ability, const Point2D& point, bool queued_command = false) override;
+    void UnitCommand(const Tags& tags, AbilityID ability, const Tag target_tag, bool queued_command = false) override;
 
     void ToggleAutocast(Tag unit_tag, AbilityID ability) override;
-    void ToggleAutocast(const std::vector<Tag>& unit_tags, AbilityID ability) override;
+    void ToggleAutocast(const Tags& unit_tags, AbilityID ability) override;
 
     void SendChat(const std::string& message, ChatChannel channel) override;
 
-    const std::vector<Tag>& Commands() const override;
+    const Tags& Commands() const override;
 
     void SendActions() override;
 
-    std::vector<Tag> commands_;
+    Tags commands_;
 };
 
 ActionImp::ActionImp(ProtoInterface& proto, ControlInterface& control) :
@@ -51,7 +57,7 @@ SC2APIProtocol::RequestAction* ActionImp::GetRequestAction() {
     return request_actions_->mutable_action();
 }
 
-const std::vector<Tag>& ActionImp::Commands() const {
+ const Tags& ActionImp::Commands() const {
     return commands_;
 }
 
@@ -81,11 +87,11 @@ void ActionImp::SendActions() {
 }
 
 void ActionImp::ToggleAutocast(Tag unit_tag, AbilityID ability) {
-    std::vector<Tag> tags = { unit_tag };
+    Tags tags = { unit_tag };
     ToggleAutocast(tags, ability);
 }
 
-void ActionImp::ToggleAutocast(const std::vector<Tag>& unit_tags, AbilityID ability) {
+void ActionImp::ToggleAutocast(const Tags& unit_tags, AbilityID ability) {
     SC2APIProtocol::RequestAction* request_action = GetRequestAction();
     SC2APIProtocol::Action* action = request_action->add_actions();
     SC2APIProtocol::ActionRaw* action_raw = action->mutable_action_raw();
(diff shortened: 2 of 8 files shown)


src/sc2api/sc2_agent.cc
```diff
@@ -122,65 +128,131 @@ void ActionImp::SendChat(const std::string& message, ChatChannel channel) {
...
+    UnitCommand(tag, ability, target_tag, queued_command);
 }
 
 void ActionImp::UnitCommand(const Units& units, AbilityID ability, bool queued_command) {
+    std::vector<Tag> tags;
+    for (auto unit : units) {
+        if (!unit) continue;
+        tags.push_back(unit->tag);
+        UnitCommand(tags, ability, queued_command);
+    }
```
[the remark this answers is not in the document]
Nickrader: It seems to me that `std::copy_if` is unable to insert `unit->tag` during it's operation?
https://en.cppreference.com/w/cpp/algorithm/copy
> The expression pred(v) ... must not modify v.

At first I was trying to std::copy_if(units.begin(), units.end(), std::back_inserter(tags), is_valid());
And in the predicate is_valid(), I was playing with lambdas and after yelling at my compiler, realized my error.

At the moment, I am doing some review of lambdas and passing a struct to output a data member.  I'll give it another look today and see if I can more accurately define my conceptual deficiencies.

tests/test_unit_command.cc
```diff
@@ -1703,8 +1703,327 @@ namespace sc2 {
         }
     };
 
+    //
+    // TestNoTarget_Tag
+    //
+
+    class TestNoTarget_Tag : public TestUnitCommandNoTarget {
```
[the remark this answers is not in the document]
Nickrader: Yes, I think we can extend the existing tests as you suggest.  I think I latched on to the first solution that presented itself to me, but the tests logic becomes clearer to me the more I play with it.  It was good practice though ;)

src/sc2api/sc2_agent.cc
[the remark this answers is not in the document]
alkurbatov: Ah, I forgot that we need to put Tag, not Unit. Well, in this case it is ok to use current approach.
It is possible to do such action by combining filter and transform (also boost provides interesting capabilities), but this solution is relatively complex for such simple task.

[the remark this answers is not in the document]
Nickrader: It was good practice trying to apply std library, since I have little experience with it, and it forced me to start attempting to use reference materials.

tests/test_unit_command.cc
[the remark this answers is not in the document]
Nickrader: With current approach that UnitCommand(Unit) will eventually call UnitCommand(Tag), does the test need to explicitly call UnitCommand(Tag) or is it sufficient that it gets called in the logic?

In `test_unit_command_common.cc` the function `IssueUnitCommand(ActionInterface*)` is setup to take Unit* or Units overload.
``` 
293        if (test_units_.size() > 1) {
294            act->UnitCommand(test_units_, test_ability_, target_unit_);
295        }
296        else {
297            act->UnitCommand(test_unit_, test_ability_, target_unit_);
298        }

// *.h
 24       const Unit* test_unit_ = nullptr;
 25       Units test_units_;
```

I think that was part of the reason I had originally went with class inheritance.  I could not think of an elegant way to change the `test_units_` data member type or substitute in another type, without recreating the function.

It makes implementation simpler and less code duplication if the Tag overload doesn't have to be explicitly called.  In that case, it would simply require making sure `TestUnitCommandTargetingUnit` , `TestUnitCommandTargetingPoint` , and `TestUnitCommandNoTarget` get used at least once as Units and Unit*.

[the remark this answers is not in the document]
alkurbatov: Speaking about best practices: we should test both versions of `UnitCommand` independently because methods could be much more complex or different in the future (consider e.g. addition of another entity which will issue request action with unit or with tag). Thus, I suggest to assume in the tests that we don't know which method calls which inside.

Speaking about `test_units_` I would try the following:
a. Take list of units.
b. Give the whole list some command (e.g. move or hold).
c. Convert the list to list of tags.
d. Give the new list another command.

You don't have to store additional list in the class, all we need is to check that the command was issued and executed.

Nickrader (commented):
1.  I couldn't figure out where I would move initialization into constructor in test_unit_command.cc
2. I couldn't figure out how to do tests without deriving from base class.  I don't completely understand the test logic.  I keep getting hung up on test_units_ being defined as type Unit and begin called in the OnStep function.

src/sc2api/sc2_agent.cc
```diff
@@ -122,66 +128,124 @@ void ActionImp::SendChat(const std::string& message, ChatChannel channel) {
...
+    SC2APIProtocol::ActionRawUnitCommand* tag_command = action_raw->mutable_unit_command();
+
+    tag_command->set_ability_id(ability);
+    tag_command->set_target_unit_tag(target_tag);
+    tag_command->set_queue_command(queued_command);
+    tag_command->add_unit_tags(tag);
 }
 
-void ActionImp::UnitCommand(const Units& units, AbilityID ability, const Unit* target, bool queued_command) {
+void ActionImp::UnitCommand(const std::vector<Tag>& tags, AbilityID ability, bool queued_command) {
```
alkurbatov, line 208: I think we should introduce `Tags` entity (e.g. by using typedef) similarly to `Units` and use it everywhere.

tests/test_unit_command.cc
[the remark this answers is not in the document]
alkurbatov: Ok, lets drop the new tests and extend the existing Units command test if needed (e.g. by adding new commands or trying to pass nullptrs). Should be ok as we use the new functionality below.

alkurbatov (changes requested):
Hey, sorry for the delay, was pretty busy last months.

Than you, the patch became better.
I suggest to do several minor improvements and accept the patch.

Nickrader:
No worries.  I can imagine life got busy.  I'll have another go at this.

alkurbatov:
Hey @Nickrader do you have any plans to finish this PR?
Currently we have lots of changes done and might have more to come so I want merge this PR before conflicts happen. If you are busy, I can finish your PR, this is totally fine.

Nickrader:
Thanks for the heads up.  I got distracted playing with Suvorov-bot and a bit lazy.  Looks like I was able to rebase it locally no conflicts yet.  I'll make a point of finishing it or getting to a standstill this weekend.  The tests may make more sense now after playing with suvurov-bot, hopefully.

Nickrader:
**by adding new commands or trying to pass nullptrs**
I can override IssueUnitCommand() in an existing test instead of creating a new class, took me awhile to see that one.  I seem to be making some headway, my new goal is to have something to show by this weekend.

Nickrader:
It seems I had to update the submodule after pulling the new changes from master, to be able to run the tests locally?  I couldn't figure out how to remove those for the push.  Besides the Git troubles I think I have come closer to what you wanted.  I don't know if I should change by hand where it committed the subproject?
Feel free to do as you like with anything, or let me know if I should do more.

alkurbatov:
Now we don't include submodules but use embedded cmake functionality. As a result, all the dependecies are put inside the build folder. So, please remove submodules and the contrib folder from the PR.

src/sc2api/sc2_unit.cc
```diff
@@ -14,6 +14,16 @@ bool Unit::IsBuildFinished() const {
     return build_progress >= 1.0f;
 }
 
+Tags GetTagsFromUnits(const Units& units) {
+    Tags tags;
+    for (const auto& unit : units) {
```
alkurbatov, line 19: There is no need in reference since sc2::Unit is a pointer which is a kind of int.

src/sc2api/sc2_unit.cc
```diff
@@ -14,6 +14,16 @@ bool Unit::IsBuildFinished() const {
...
+        if (!unit) continue;
+        tags.push_back(unit->tag);
```
alkurbatov, line 21: I believe there is no need in checking for nullptr. This is much better to crash as nullptr in units list means some weird error in program.

Also, please try this solution, it allows to use algorithms from standard library and could be faster:
https://stackoverflow.com/questions/59840478/using-back-inserter-to-create-vector-from-single-member-of-struct

include/sc2api/sc2_unit.h
```diff
@@ -218,8 +218,11 @@ class Unit {
 };
 
 typedef std::vector<const Unit*> Units;
+typedef std::vector<Tag> Tags;
 typedef std::unordered_map<Tag, size_t> UnitIdxMap;
 
+Tags GetTagsFromUnits(const Units& units);
```
alkurbatov, line 224: Please rename this function to `ConvertToTags`. This is minor style issue, the new name is more consistent with other parts of the code.

tests/test_unit_command.cc
```diff
@@ -18,9 +18,104 @@ namespace sc2 {
...
+            agent_->Debug()->SendDebug();
+        }
+
         void SetTestTime() override {
-            wait_game_loops_ = 50;
+            wait_game_loops_ = 150;
         }
+
+        void IssueUnitCommand(ActionInterface* act) {
+
```
alkurbatov, line 33: Style: Please drop the empty string at the beginning.

tests/test_unit_command.cc
```diff
@@ -18,9 +18,104 @@ namespace sc2 {
...
+                    break;
+                default:
+                    break;
+                }
+
+                ability_command_sent_ = true;
+            }
+        }
+
+            void TestTagsTargetingPoint(ActionInterface * act) {
```
alkurbatov, line 73: Please, be consistent with the style. Here and in the other places.

tests/test_unit_command.cc
```diff
@@ -1800,6 +1892,10 @@ bool TestUnitCommand(int argc, char** argv) {
         CreateParticipant(sc2::Race::Terran, &bot),
     });
 
+    //// Realtime can cause OnUnitIdle() error in some current tests
+    //// "Error: Unit did not trigger OnUnitIdle as expected."
+    //coordinator.SetRealtime(true);
```
alkurbatov, line 1897: Please drop the commented code.

tests/test_unit_command.cc
```diff
@@ -18,9 +18,104 @@ namespace sc2 {
...
+                if (!second_pass && ability_command_sent_) {
+                    test_unit_tag_ = test_units_tags_.front();
+                    orders_verified_ = false;
+                }
+
+                uint32_t current_game_loop = agent_->Observation()->GetGameLoop();
+                order_on_game_loop_ = current_game_loop + 10;
+
+                switch (test_tags_counter_) {
+                case 0: TestTagsTargetingPoint(act);
```
alkurbatov, line 59: I think this should be separate test classes sharing the same setup code. The current implementation make the code much more complex (as you need to handle these first and second passes) while it is much better to keep the tests as simple as possible.

alkurbatov (changes requested):
Thank you, the src and include parts are almost good. The tests should be be improved a bit.

tests/test_unit_command.cc
```diff
@@ -13,14 +13,108 @@ namespace sc2 {
...
+                        break;
+                    default:
+                        break;
+                }
+
+                ability_command_sent_ = true;
+            }
+        }
+
+            void TestTagsTargetingPoint(ActionInterface* act) {
```
alkurbatov, line 77: Please, check the indentation. Seems that it is a bit extra.

alkurbatov: Might be also tabs or spaces.

tests/test_unit_command.cc
```diff
@@ -13,14 +13,108 @@ namespace sc2 {
...
+                    orders_verified_ = false;
+                    ability_command_sent_ = false;
+                    ++test_tags_counter_;
+                    return;
+                }
+
+                uint32_t current_game_loop = agent_->Observation()->GetGameLoop();
+                order_on_game_loop_ = current_game_loop + 10;
+
+                switch (test_tags_counter_) {
```
alkurbatov, line 50: Thank you! That's much better comparing to the previous approach.

alkurbatov:
LGTM except the indentation issue.

tests/test_unit_command.cc
Nickrader: Yeah, I've become dependent on auto-formatting, and don't have it setup right for this repo, so we are seeing my lack of doing it myself during my learning.  LOL.  Yup, I just need to follow the braces.

alkurbatov (approved):
Thank you!

Nickrader:
You're welcome, thanks for the help.

PR #113, opened 2022-08-02
build: msvc cmake cleanup

jlknauer:
My original plan for work on the project was to work on supporting cross-compilation from Linux to Windows in a WSL2 context (using precompiled libraries, xwin for the windows SDK, and Clang) for the BlankBot repository. This would enable users to download a zipped cross-compilation package published by cpp-sc2 and easily build bots from cpp-sc2 in the WSL2 context while using the SC2 executable installed in the windows filesystem.

However, the underlying CMake structure and configuration carried forward from the s2client-api introduces a variety of unexpected side effects which interfered with my original work. I opted to take the time to do a first pass at adjusting the CMake configuration (within reason) to facilitate future development.

There are more changes which *can* be made (e.g. replacing file globbing with explicit source and header file naming, taking a closer look at the remaining compile flags, being more specific to STATIC libraries as dynamic linking was fundamentally broken and should be disabled outright), but I believe this is a good starting point for the purpose of:

- Isolating build logic for each library
- Leveraging CMake's target inheritance scope for includes and linking to simplify the project's configuration
- Replacing old CMake hacks with proper abstractions
- Removing unnecessary and/or repetitious CMake configuration
- General style changes for consistency

I have tested the configuration changes for Debug and Release for:

- Visual Studio 17 2022
- Visual Studio 15 2019
- Visual Studio 15 2017 Win64
- gcc/g++ 9.4.0

I would hope that should this PR be accepted, project documentation can be updated to include support for Visual Studio 17 2022.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.15)
+cmake_policy(VERSION 3.15)
 
 project(cpp-sc2)
 
@@ -35,44 +36,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELWITHDEBINFO ${PROJECT_BINARY_DIR}/bin)
 # Build with c++14 support.
 set(CMAKE_CXX_STANDARD 14)
 
-# Disable default libraries in visual studio
-if (MSVC)
-    # Default to statically-linked runtime.
-    if("${MSVC_RUNTIME}" STREQUAL "")
-        set(MSVC_RUNTIME "static")
-    endif()
-
-    # Set compiler options.
-    set(variables
-        CMAKE_C_FLAGS_DEBUG
-        CMAKE_C_FLAGS_MINSIZEREL
-        CMAKE_C_FLAGS_RELEASE
-        CMAKE_C_FLAGS_RELWITHDEBINFO
-        CMAKE_CXX_FLAGS_DEBUG
-        CMAKE_CXX_FLAGS_MINSIZEREL
-        CMAKE_CXX_FLAGS_RELEASE
-        CMAKE_CXX_FLAGS_RELWITHDEBINFO
-    )
-    if(${MSVC_RUNTIME} STREQUAL "static")
-        message(STATUS
-          "MSVC -> forcing use of statically-linked runtime."
-        )
-        foreach(variable ${variables})
-            if(${variable} MATCHES "/MD")
-                string(REGEX REPLACE "/MD" "/MT" ${variable} "${${variable}}")
-            endif()
-        endforeach()
-    else()
-        message(STATUS
-            "MSVC -> forcing use of dynamically-linked runtime."
-        )
-        foreach(variable ${variables})
-            if(${variable} MATCHES "/MT")
-                string(REGEX REPLACE "/MT" "/MD" ${variable} "${${variable}}")
-            endif()
-        endforeach()
-    endif()
-endif ()
+# Use statically linked runtime
+set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
 # Allow creating filters for projects in visual studio.
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,13 +80,19 @@ set_target_properties(libprotoc PROPERTIES FOLDER contrib)
 set_target_properties(protoc PROPERTIES FOLDER contrib)
 set_target_properties(ipv6-parse PROPERTIES FOLDER contrib)
 
+# civetweb's cmake does not set debug postfix
+set_target_properties(civetweb-c-library PROPERTIES DEBUG_POSTFIX "d")
+
 if (MSVC)
+    # Setup MSVC parallelized builds
+    add_compile_options($<$<CXX_COMPILER_ID:MSVC>:/MP>)
+
     set_target_properties(libprotobuf libprotobuf-lite libprotoc protoc PROPERTIES COMPILE_FLAGS "/W0")
     set_source_files_properties(${libprotobuf_files} PROPERTIES COMPILE_FLAGS "/W0")
     set_source_files_properties(${protobuf_SHARED_OR_STATIC} PROPERTIES COMPILE_FLAGS "/W0")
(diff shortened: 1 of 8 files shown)


src/cppsc2_interface/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Single target interface library for bots to link against
+add_library(cppsc2_interface INTERFACE)
+
+target_link_libraries(cppsc2_interface
```
alkurbatov, line 4: Could you please briefly describe what's the idea behind creating this interface as a separate library?
From what I've seen in the modern c++ libs approach with targets (e.g. https://git.mts.ai/ai/voiceplatform/audio-transcoder/-/blob/develop/cmake/FindgRPC.cmake#L205) .

CMakeLists.txt
```diff
@@ -35,44 +36,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELWITHDEBINFO ${PROJECT_BINARY_DIR}/bin)
...
-        )
-        foreach(variable ${variables})
-            if(${variable} MATCHES "/MT")
-                string(REGEX REPLACE "/MT" "/MD" ${variable} "${${variable}}")
-            endif()
-        endforeach()
-    endif()
-endif ()
+# Use statically linked runtime
+set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
```
alkurbatov, line 39: Shouldn't the actual library depend on particular configuration? Cause here we use Debug in any case, even if Release was asked.

alkurbatov (commented):
Looks good to me, but I have couple of questions.

src/cppsc2_interface/CMakeLists.txt
jlknauer: Take the current CMake configuration for [BlankBot's src dir](https://github.com/cpp-sc2/blank-bot/blob/main/src/CMakeLists.txt):
```
# Create the executable.
add_executable(BlankBot ${BOT_SOURCES})
target_link_libraries(BlankBot
    sc2api sc2lib sc2utils sc2protocol civetweb libprotobuf
)
```

Some of these targets are cleaned up already (in particular protobuf through sc2protocol, and then alongside civetweb into sc2api, which would then consolidate the configuration to:
```
# Create the executable.
add_executable(BlankBot ${BOT_SOURCES})
target_link_libraries(BlankBot
    sc2api sc2lib sc2utils
)
```

I think this is fine, but does not accurately convey the intent of the `cpp-sc2` project usage. While `sc2api`, `sc2lib` and `sc2utils` are three separate compilation units, there are few independent use cases (one being pure bot library code, such as `sc2lib` which only needs to link to `sc2api`). But for an individual who just wants to resolve all dependencies required for a successful bot build from the boilerplate template, `cppsc2_interface` conveys the intent of these three libraries being the minimal targets required to use any of the API and library code.

Plus, there is no compilation overhead as there is no `cppsc2_interface` library artifact being produced as it is `INTERFACE` without sources (I've read about `INTERFACE` libraries generating other targets with custom commands, but that's outside of my knowledge and scope for this change).

I also have my own thoughts as to whether `sc2lib` even belongs in `cpp-sc2` as a dedicated library (I really think bot library code like this should be pulled into its own submodule, but the implementation released is so sparse and really exists to serve as an extension of the `examples`, providing users an idea of how to separate their own bot implementation from the project's).

A bit of a roundabout, but in summary:

- **Yes**, `cppsc2_interface` does not need to exist. Always linking against `sc2api`, `sc2utils`, and `sc2lib` when used is acceptable.
- `cppsc2_interface` conveys project intent, as to build a bot using `cpp-sc2` these three (but in reality only two are truly necessary, but the docs walkthrough requires `sc2lib` so we pull it in anyway) libraries are necessary.
- Especially in the case of pulling in `cpp-sc2` as a submodule (e.g. BlankBot) the library user shouldn't need to figure out what targets are necessary to use the project to its fullest. Providing a single interface target keeps things simple, for little build cost (I'm not in a position to quantify how many additional CPU cycles CMake will need to dedicate towards resolving the inheritance of one additional node in the tree)

CMakeLists.txt
jlknauer: Not sure if you've misread the generator expression or I'm misunderstanding your concern:
`"MultiThreaded$<$<CONFIG:Debug>:Debug>" will expand to `MultiThreaded` if not `Debug` (so any flavor of `Release`), and `MultiThreadedDebug` if `Debug`. 

To go into more detail of why the old implementation needed to be replaced:

The previous implementation was troublesome as if I moved up the CMake version without setting [CMAKE_MSVC_RUNETIME_LIBRARY](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html), I would get the unpleasant default multithreaded behavior. Symbols between my `BlankBot main.cpp` code would mismatch against the `cpp-sc2` compiled libraries and nothing would work. The regex implementation also was post-processing compilation flags, which was understandable at the time the commit was made as the runtime library abstraction simply wasn't available in CMake at the time. But now there's a definitive means of setting the runtime configuration.

src/cppsc2_interface/CMakeLists.txt
alkurbatov: I see, thank you for the good description!
From my point of view there is no need in empty dir with single cmake file, but it is ok to keep it this way.

What do you think about renaming it into something like `cpp_sc2::cpp_sc2`? Just to use the same convention as other libraries e.g. fmt or gRPC.

CMakeLists.txt
alkurbatov: Ah, this is something new in cmake syntax I am not aware of. Thank you, this part looks good.

src/cppsc2_interface/CMakeLists.txt
jlknauer: I think cpp_sc2 will be sufficient. `::` syntax is reserved for [imported or alias target names](https://cmake.org/cmake/help/latest/policy/CMP0028.html#policy:CMP0028).

The empty directory is unnecessary. I have no issues pushing the `INTERFACE` logic up into the `src` level build. Changes have been pushed.

jlknauer:
As this is my first time contributing here, for PR merges do you want the PR squashed prior to merge?

alkurbatov:
I prefer the rebase especially when commits have good description like in your PR.

As for the last commit, I would do it inside one of the previous (i.e. rebase, amend, force push) but this is minor so we can left it as is.

alkurbatov (approved):
Thank you!
Looks good to me.

PR #114, opened 2022-08-05
build: disable ipv6-parse

jlknauer:
ipv6-parse was introduced as a submodule in [0], but no source files or
dependencies include headers from it. The sole source file which is
likely to have referenced it is the example project source file
bot_mp_ipv6.cc, which has never been added to the project nor enabled
(it has been commented out since it was merged in [1]).

Refs:
    [0] 7286ea4000820a90bc5335b16e881ebec849365f
    [1] https://github.com/cpp-sc2/cpp-sc2/commit/7286ea4000820a90bc5335b16e881ebec849365f#diff-940e5356fa1137e646ea135a6f68cacbfad4fe7124c3a69163468f588acf9283R43

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "protocol"]
 	path = protocol
 	url = https://github.com/Blizzard/s2client-proto
-[submodule "contrib/ipv6-parse"]
-	path = contrib/ipv6-parse
-	url = https://github.com/jrepp/ipv6-parse
 [submodule "contrib/civetweb"]
 	path = contrib/civetweb
 	url = https://github.com/cpp-sc2/civetweb.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,9 +51,6 @@ set(CIVETWEB_ENABLE_IPV6 ON CACHE BOOL "" FORCE)
 # Don't build civetweb with sanitizers
 set(CIVETWEB_ENABLE_ASAN OFF CACHE BOOL "" FORCE)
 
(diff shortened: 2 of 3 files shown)


jlknauer:
Just to open the discussion of whether or not to keep the submodule in the project, so my commit only disables it in the build.

alkurbatov:
From my point of view ipv6 should be removed together with that ipv6 bot example. I kept this submodule only because it was unclear in the past whether we will use ipv6 on AIArena or not. Currently I am 100% sure we will not need ipv6, so leads just drop it.

alkurbatov (changes requested):
Please remove the submodule too.

alkurbatov:
> From my point of view ipv6 should be removed together with that ipv6 bot example. I kept this submodule only because it was unclear in the past whether we will use ipv6 on AIArena or not. Currently I am 100% sure we will not need ipv6, so lets just drop it.

Another point here is that the examples could be (or rather should be) moved outside of this repo. By some reason many people don't even know that we have them, so separate repo pinned at the top could be more informative :)

jlknauer:
`.gitmodules` didn't get staged on first pass so had to rebase that in and force.

> Currently I am 100% sure we will not need ipv6, so lets just drop it.

I can already see the ping in the discord tomorrow for migrating to ipv6 addresses. But yes seems very unlikely. Not that ipv6-parse took any considerable build time - but no point in keeping it around as it will just confuse anyone else trying to understand the build dependency tree.

alkurbatov (approved):

PR #117, opened 2022-08-10
fix: SDL Convert BGR to RGB on Linux

danielvschoor:
On Linux, the binary produces the render data in a BGR format and on Windows (and maybe Mac) it's produced as RGB. This PR changes the SDL surface to render BGR on Linux

### DIFF
--- a/src/sc2renderer/sc2_renderer.cc
+++ b/src/sc2renderer/sc2_renderer.cc
@@ -135,8 +135,13 @@ void Matrix8BPPPlayers(const char* bytes, int w_mat, int h_mat, int off_x, int o
 void ImageRGB(const char* bytes, int width, int height, int off_x, int off_y) {
     assert(renderer_);
     assert(window_);
-
+#ifdef linux
+    SDL_Surface *surface = SDL_CreateRGBSurfaceWithFormatFrom((void *) bytes, width, height, 24, 3 * width,
+                                                              SDL_PixelFormatEnum::SDL_PIXELFORMAT_RGB24);
+#else
     SDL_Surface* surface = SDL_CreateRGBSurfaceFrom((void*)bytes, width, height, 24, 3 * width, 0xFF0000, 0x00FF00, 0x0000FF, 0x000000);
+#endif
+
     SDL_Texture* texture = SDL_CreateTextureFromSurface(renderer_, surface);
     SDL_FreeSurface(surface);
 
--- a/src/sc2renderer/sc2_renderer.cc
+++ b/src/sc2renderer/sc2_renderer.cc
@@ -135,8 +135,8 @@ void Matrix8BPPPlayers(const char* bytes, int w_mat, int h_mat, int off_x, int o
 void ImageRGB(const char* bytes, int width, int height, int off_x, int off_y) {
     assert(renderer_);
     assert(window_);
-#ifdef linux
-    SDL_Surface *surface = SDL_CreateRGBSurfaceWithFormatFrom((void *) bytes, width, height, 24, 3 * width,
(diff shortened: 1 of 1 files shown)


alkurbatov:
Thanks! I'll check on OS X and merge if everything is ok.

src/sc2renderer/sc2_renderer.cc
```diff
@@ -135,8 +135,13 @@ void Matrix8BPPPlayers(const char* bytes, int w_mat, int h_mat, int off_x, int o
 void ImageRGB(const char* bytes, int width, int height, int off_x, int off_y) {
     assert(renderer_);
     assert(window_);
-
+#ifdef linux
+    SDL_Surface *surface = SDL_CreateRGBSurfaceWithFormatFrom((void *) bytes, width, height, 24, 3 * width,
```
alkurbatov, line 139: Style: Please, follow style used in this file. Currently pointer declaration style is different.

src/sc2renderer/sc2_renderer.cc
```diff
@@ -135,8 +135,13 @@ void Matrix8BPPPlayers(const char* bytes, int w_mat, int h_mat, int off_x, int o
 void ImageRGB(const char* bytes, int width, int height, int off_x, int off_y) {
     assert(renderer_);
     assert(window_);
-
+#ifdef linux
```
alkurbatov, line 138: Please use `#if defined(__linux__)`.
This is again mostly for style reasons as this way is used to check target system.

alkurbatov (changes requested):

alkurbatov:
Could you please provide an example image of what should be shown?
So far with both ways on OS X I see two windows: one with a game (map + screen) and another one is pure black.
I wonder which part this change affects.

danielvschoor:
> Could you please provide an example image of what should be shown?
> So far with both ways on OS X I see two windows: one with a game (map + screen) and another one is pure black.
> I wonder which of the windows this patch affects.

It should only show one screen with both the minimap and the game inside. This might be an OS X specific issue, but not sure what could cause it

danielvschoor:
image

Just one screen on Linux^^

alkurbatov:
I see. The second window spawned by the renderer example which I used to check the patch.
I'll check it again, perhaps I can spot difference in these two lines. Just curios.

Meanwhile, please squash the commits.

src/sc2renderer/sc2_renderer.cc
```diff
@@ -135,8 +135,13 @@ void Matrix8BPPPlayers(const char* bytes, int w_mat, int h_mat, int off_x, int o
 void ImageRGB(const char* bytes, int width, int height, int off_x, int off_y) {
     assert(renderer_);
     assert(window_);
-
+#if defined(__linux__)
+    SDL_Surface* surface = SDL_CreateRGBSurfaceWithFormatFrom((void *) bytes, width, height, 24, 3 * width,
```
alkurbatov, line 139: `(void *) bytes` -> `(void*)bytes`

alkurbatov:
AT least on OS X there are no visible difference in both of these strings, lets keep the idea and apply the change only for linux. Please, fix minor style issues and I'll merge.

src/sc2renderer/sc2_renderer.cc
alkurbatov: Ok, from what I've found this is a shorter form (i.e. without pixels) of the original SDL_CreateRGBSurface, so there is no need in defined guards, we should use it on all platforms or should correct the original pixels mask.

alkurbatov:
Closed in favor of #135

PR #118, opened 2022-08-16
build: cmake dependency approach refactor

jlknauer:
The current git submodule approach for project dependencies tightly
integrates subprojects as part of the main project. CMake's FetchContent
is used as a configurable alternative to explicit submodule inclusion.
Users can freely modify the dependency origin, and also facilitates
further refactors and separation of depedency build configuration
from the cpp-sc2 library CMake files.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,16 +1,16 @@
 # Function to generate an example project with extra libraries
 function (example_project_extra project_name source_code extra_libs)
-    message(VERBOSE "Adding new example project ${project_name}, sources: ${source_code}, extra_libs: ${extra_libs}")
-    file (GLOB example_project_sources ${source_code} "common/*.h" "common/*.cc")
+    message(VERBOSE "Adding new example project ${project_name},\
+            sources: ${source_code}, extra_libs: ${extra_libs}")
+    set(example_project_sources
+        ${source_code}
+        common/bot_examples.cc)
+
     add_executable(${project_name} ${example_project_sources})
 
-    target_include_directories(${project_name} PRIVATE "${PROJECT_SOURCE_DIR}/examples/common")
+    target_include_directories(${project_name} PRIVATE common)
 
     set_target_properties(${project_name} PROPERTIES FOLDER examples)
-
-	if (MSCV AND (NOT (MSVC_VERSION LESS 1910)))
-		target_link_libraries(${project_name} PRIVATE legacy_stdio_definitions)
-	endif ()
 
     target_link_libraries(${project_name} PRIVATE cpp_sc2 ${extra_libs})
 endfunction ()
@@ -20,19 +20,18 @@ function (example_project project_name source_code)
     example_project_extra(${project_name} ${source_code} "")
 endfunction ()
 
-example_project(bot_simple "bot_simple.cc")
-example_project(echo_actions "echo_actions.cc")
-example_project(replay "replay.cc")
-example_project(tutorial "tutorial.cc")
-example_project(bot_micro "bot_micro.cc")
-example_project(bot_mp "bot_mp.cc")
-example_project(intermediate_bot "intermediate_bot.cc")
-example_project(annoying_helper "annoying_helper.cc")
-example_project(proxy "proxy.cc")
-example_project(save_load "save_load.cc")
+example_project(bot_simple bot_simple.cc)
+example_project(echo_actions echo_actions.cc)
+example_project(replay replay.cc)
+example_project(tutorial tutorial.cc)
+example_project(bot_micro bot_micro.cc)
+example_project(bot_mp bot_mp.cc)
+example_project(intermediate_bot intermediate_bot.cc)
+example_project(annoying_helper annoying_helper.cc)
+example_project(proxy proxy.cc)
+example_project(save_load save_load.cc)
 
 if (BUILD_SC2_RENDERER)
-    example_project_extra(feature_layers "feature_layers.cc" sc2renderer)
-    example_project_extra(rendered "rendered.cc" sc2renderer)
+    example_project_extra(feature_layers feature_layers.cc sc2renderer)
+    example_project_extra(rendered rendered.cc sc2renderer)
 endif ()
-
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,16 +1,10 @@
-if (MSVC)
-    set(CMAKE_CXX_FLAGS_DEBUG "/ZI /Od" CACHE STRING "" FORCE)
-else ()
-    set(CMAKE_CXX_FLAGS "-fpermissive")
-endif ()
-
-add_subdirectory("sc2api")
-add_subdirectory("sc2lib")
-add_subdirectory("sc2utils")
-add_subdirectory("sc2protocol")
+add_subdirectory(sc2api)
+add_subdirectory(sc2lib)
+add_subdirectory(sc2utils)
+add_subdirectory(sc2protocol)
 
 if (BUILD_SC2_RENDERER)
-    add_subdirectory("sc2renderer")
+    add_subdirectory(sc2renderer)
 endif ()
 
 add_library(cpp_sc2 INTERFACE)
--- a/src/sc2api/CMakeLists.txt
+++ b/src/sc2api/CMakeLists.txt
@@ -1,6 +1,3 @@
-# All the source files for the sc2 api
-file(GLOB sources_sc2 "*.cpp" "*.cc" "*.h")
-
 # Inject sources specific for particular version of the game client
 file(WRITE "${PROJECT_SOURCE_DIR}/include/sc2api/sc2_typeenums.h"
 "#pragma once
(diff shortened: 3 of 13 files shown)


jlknauer:
Not looking for a merge right now - would like some feedback of the approach and next steps. This refactor can effectively eliminate keeping submodules in project, as they are fetched as part of the build configuration. There are some additional smaller improvements I can make, but the work done so far is one step short of working out packaging `cpp-sc2` through CMake so it can be imported by projects like `BlankBot` with a single fetch and configuration.

The cleanup also handled some minor carryover from switching off the SDL mirror (introduced some additional builds which are not necessary), replacing all file globbing with explicit sources (some unnecessary headers were being globbed in for compilation...), and a variety of other CMake related improvements.

Naturally there is a small tradeoff, as a clean configuration build (running `cmake -B build` for the first time) will introduce some download overhead to fetch the external subprojects, but I think the increased configuration time is negligible. Updating pins, for both development and maintenance is also simplified as it is handled in the CMake configuration instead of updating the submodule itself.

cmake/External_SDL.cmake
```diff
@@ -0,0 +1,20 @@
+message(STATUS "FetchContent: SDL")
```
alkurbatov, line 1: I think this folder should be named vendor or thirdparty to indicate that these dependencies are provided by other people/organizations. In this case we can safely drop the "External" prefix from the names of macros as the folder name becomes a kind of namespace.

src/CMakeLists.txt
```diff
@@ -1,7 +1,5 @@
-if (MSVC)
-    set(CMAKE_CXX_FLAGS_DEBUG "/ZI /Od" CACHE STRING "" FORCE)
-else ()
-    set(CMAKE_CXX_FLAGS "-fpermissive")
+if (NOT MSVC)
+    add_compile_options(-fpermissive)
```
alkurbatov, line 2: Lets drop `-fpermissive` too. I doubt there is any need in such tricks. If it still exists, better to fix that code.

src/sc2api/CMakeLists.txt
```diff
@@ -9,26 +6,33 @@ file(WRITE "${PROJECT_SOURCE_DIR}/include/sc2api/sc2_typeenums.h"
...
+    sc2_replay_observer.cc
+    sc2_score.cc
+    sc2_server.cc
+    sc2_unit_filters.cc
+    sc2_unit.cc
+    "${PROJECT_SOURCE_DIR}/src/sc2api/typeids/sc2_${SC2_VERSION}_typeenums.cpp")
+
+add_library(sc2api STATIC ${sc2api_sources})
+
+target_include_directories(sc2api PUBLIC "${PROJECT_SOURCE_DIR}/include")
```
alkurbatov, line 31: This change cause small degradation: now we include all version headers while we need only two of them.
Not critical.

cmake/External_SDL.cmake
alkurbatov: Also, we may include additional dependencies (e.g. libfmt as a header) so the folder will be used anyway.

alkurbatov:
Regarding the proposal:
I am ok with removal of files globbing.

I am mostly ok with the fetch approach but have some questions:

1. Is it really more familiar to other people? Honestly I haven't seen it widely used.
2. How does it help in packaging?
3. What if I change the hash after the full build? Should I recreate the build folder and fetch everything from scratch?

jlknauer:
[Here](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#guide:Using%20Dependencies%20Guide) is CMake's documentation outlining the high level approaches available for managing dependencies, but it can be summarized to:
- `find_package` for pre-built or installed packages (so especially useful for containerized images where you can reliably download and configure the dependencies, and *then* build the project).
- `FetchContent`, which is an abstraction over `ExternalProject` - but will download the dependencies at configuration time instead of build time.
- Both can be used together with the `FIND_PACKAGE_ARGS` setting for `FetchContent` to run `find_package` for `NAMES` before downloading the dependency.

So to answer your questions in order:

> Is it really more familiar to other people? Honestly I haven't seen it widely used.

`FetchContent` was added in [version 3.11](https://github.com/Kitware/CMake/tree/v3.11.0), which released a little under 4 years ago. The only alternative at the time for ("superbuilds")[https://www.kitware.com/cmake-superbuilds-git-submodules/] strategy would be to use `ExternalProject`, which has the drawback of pulling in dependencies at build time. 

In a perfect world every project using CMake would configure export and install targets, and `find_package` would be the standard. `civetweb`, `SDL` and `protobuf` all actually define these targets, so `cpp-sc2` could easily enough have the user install these projects separately - but as we are currently pinning to not only git hashes but also a mixture of forked and original repositories, I would hesitate to go all in on the `find_package` route. If we were pinning to specific packaged releases of a dependency it would be more appropriate, but this leads into the next question.

> how does it help in packaging

As described in the CMake dependency guide, there is more than one way to include dependencies in a project. Currently, dependencies are maintained at git submodules - so outside of the context of the build system. CMake has no explicit knowledge of the dependencies outside of the manually set `add_subdirectory` paths we maintain. The side-effects are already present in the build configuration, such as manual checks in `sc2protocol`:
```
if ("${proto_files}" STREQUAL "")
    message(FATAL_ERROR "Missing root protocol submodule.\nRemove your protocol directory and run:\ngit submodule update --recursive --init")
endif ()
```
Relying on git submodules to manage dependencies for the build requires the inclusion of git submodule logic, whereas with `FetchContent` we can isolate dependency management to the build itself, as well as provide a more configurable interface.

> What if I change e.g. SDL2 hash after the full build? Should I recreate the build folder and fetch everything from scratch?

CMake looks for changes between the build tree files and the configuration. Using the example of file globbing, adding a file to a source directory would not be detected by CMake globbing with `CONFIGURE_DEPENDS` on `MSVC` as it isn't supported, and would necessitate a full rebuild.

`FetchContent_Declare` and `FetchContent_MakeAvailable` have [well defined](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare) procedures and expected behaviors. Changing the git hash for a dependency will download and update the build tree sources *only for the changed dependency*. All `FetchContent` commands will always run on configuration builds, but `FetchContent_MakeAvailable` checks to see if the dependency requirements are already met (if it is already available and up to date, and in the correct directories, **do nothing**).

This is where I think the tradeoffs for `FetchContent` are most powerful, as the overhead is minimal and now all dependency logic is restricted to CMake. Users can freely change dependency repos (e.g. between forks) and hashes without needing to step into submodules and the edge-case headaches associated with it. Returning the the "perfect world" example, projects would publish/maintain a minimal repository branch / package for download which only contains the sources, headers, and CMake configuration required to include as a subproject using `FetchContent`. Just as with gitsubmodules, `FetchContent` unfortunately will still drag in all documentation, miscellaneous scripts, etc. from the subproject, but we at least still have the configuration capability to turn off all unnecessary target builds (tests, shared libraries).

I hope this answers your questions and sufficiently makes the case for using `FetchContent` over git submodules. There is still some target level work I have left to do (after reading more documentation, we *should* be able to use properly namespaced targets from dependencies such as civetweb::civetweb-c-library, but I have not tested this yet).

alkurbatov:
Thank you! Now the subject is much more clearer.

>civetweb, SDL and protobuf all actually define these targets, so cpp-sc2 could easily enough have the user install these projects separately 
Oh well, I had lots of issues with people who have no idea how to install e.g. boost, especially on Windows. The submodules are more or less common now so usually don't rise questions. So I would prefer to keep the current approach (i.e. with fetching) in any case. Partly because it doesn't force people to install dependencies manually (and come to me with questions like where one can find protobuf and how to compile it :) ), partly because I want to have convenient access to the sources especially if I want to patch something or experiment with different version without altering the packages in the system.

Another use case for this fetching mechanic is patching of the dependencies. Currently we use a patched fork of civetweb with several patches from Blizzard. While I have no idea do we really need to keep them, I would prefer to explicitly include these patches in cpp-sc2, apply during checkout and drop that fork.

Thus, green light from me :)

jlknauer:
Sounds good. I too prefer `FetchContent` over `find_package` in this case. If we were publishing docker images with everything cleanly packaged, then it would make more sense (and would really tighten up the includes as we wouldn't need to target include subproject header files, as they would all be exported to the namespaced target like `civetweb::civetweb-c-library`). But CMake is a realm of trade-offs, so just trying to make the best of what works for this project.

I'll just be applying some additional adjustments + documentation updates, and the submodules can then be removed once everything is verified.

As an aside on `FetchContent`, I can't find the original mailing list thread but it is my understanding the author of the module - Craig Scott - implemented `FetchContent` precisely because he didn't like working with git submodules and `ExternalProject` on very large projects (he described managing 40+ dependencies and needing to work within the dependencies in parallel to the main project while testing), and the module seems to be the logical resolution to the problem.

alkurbatov:
Yeah, the management overhead is smaller with `FetchContent`. Totally agree on this.

examples/CMakeLists.txt
```diff
@@ -1,17 +1,17 @@
 # Function to generate an example project with extra libraries
 function (example_project_extra project_name source_code extra_libs)
     message(VERBOSE "Adding new example project ${project_name}, sources: ${source_code}, extra_libs: ${extra_libs}")
-    file (GLOB example_project_sources ${source_code} "common/*.h" "common/*.cc")
+    # file (GLOB example_project_sources ${source_code} "common/*.h" "common/*.cc")
```
alkurbatov, line 4: Please drop the commented code.
Also, I would suggest to get rid of this generator function in favor of dedicated makefile per project (not in this PR).

tests/CMakeLists.txt
```diff
@@ -1,21 +1,30 @@
...
+    test_actions.cc
+    test_app.cc
+    test_feature_layer_mp.cc
+    test_feature_layer.cc
+    test_framework.cc
+    test_movement_combat.cc
+    test_multiplayer.cc
+    test_observation_interface.cc
+    test_performance.cc
+    test_rendered.cc
```
alkurbatov, line 14: I believe `test_rendered.cc` shouldn't be included here if we are compiling disabled `BUILD_SC2_RENDERER`, but the original code has this issue too.

jlknauer:
So I did a final pass on my changes this weekend alongside working on cross compilation support for Windows under WSL2 using the `x86_64-w64-mingw32` toolchain. Naturally that work will be presented in a different PR, but some additional cleanup to the `sc2protocol` configuration has been included to facilitate the upcoming changes.

I have also normalized the usage of quotes for CMake variable expansion and setting according to the consolidated tips found in [this thread](https://stackoverflow.com/questions/35847655/when-should-i-quote-cmake-variables). The big takeaway is quotes are most important for expanding variables which may have *spaces* or *colons* - so any paths with variable expansion *should* be quoted. `cpp-sc2` generally does not concern itself with colons for project configuration.

I'm content with the PR as it is (most recent rebase has dealt with the outstanding issues), so I have nothing else to add.

alkurbatov (approved):
Thank you! Quite a piece of work.

PR #119, opened 2022-08-22
Wsl2 support

jlknauer:
Opening a PR to demonstrate current approach, but not ready for merge.

TODOs include:
- Build documentation updates
- A central archive of precompiled `pb.cc` and `pb.h` protos for Windows to fetch from
- Some cleanup in the `CMakeLists.txt` file for sc2protocol, as it is currently just copying the downloaded files into `generated/s2clientprotocol`. Should conditionally use the downloaded files correctly, but requires some refinement on how paths are being set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,18 @@
 cmake_minimum_required(VERSION 3.15)
 
+# Skip generating proto files with Protobuf compiler and use prebuilt files
+# Will still build Protobuf as libprotobuf is required for linking
+option(WSL2_CROSS_COMPILE "Building on WSL2 for Windows" OFF)
+
+# Only valid OS configuration should be in WSL2 context
+# Should not work wtih WIN32, APPLE, etc.
+if (WSL2_CROSS_COMPILE AND UNIX AND NOT APPLE)
+    set(CMAKE_TOOLCHAIN_FILE cmake/toolchain/x86-64-w64-mingw32.cmake)
+elseif (WSL2_CROSS_COMPILE)
+    message(FATAL_ERROR "Invalid environment.\
+                        WSL2_CROSS_COMPILE=ON for WSL only")
+endif ()
+
 project(cpp-sc2)
 
 include(FetchContent)
@@ -33,7 +46,9 @@ set(LIBRARY_DEBUG_POSTFIX d)
 set(CMAKE_DEBUG_POSTFIX ${LIBRARY_DEBUG_POSTFIX})
 
 # External subprojects
-list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/thirdparty/cmake")
+list(APPEND CMAKE_MODULE_PATH
+    "${PROJECT_SOURCE_DIR}/thirdparty/cmake"
+    "${PROJECT_SOURCE_DIR}/cmake/assets")
 
 # Build with c++14 support.
 set(CMAKE_CXX_STANDARD 14)
@@ -55,6 +70,10 @@ endif ()
 
 if (BUILD_API_TESTS)
     add_subdirectory(tests)
+endif ()
+
+if (WSL2_CROSS_COMPILE)
+    include(protos)
 endif ()
 
 # External subproject depedencies
--- a/cmake/assets/protos.cmake
+++ b/cmake/assets/protos.cmake
@@ -1,0 +1,9 @@
+message(STATUS "FetchContent: generated protos for Windows")
+
+FetchContent_Declare(
+    protos
+    URL https://storage.yandexcloud.net/cpp-sc2/wsl2/generated-s2clientprotocol-db14236-protobuf-1a74ba4.zip
+    URL_HASH MD5=346e3d8127f2da6cea4900465183314b
+    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+)
+FetchContent_MakeAvailable(protos)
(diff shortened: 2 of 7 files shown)


cmake/assets/protos.cmake
```diff
@@ -0,0 +1,9 @@
+message(STATUS "FetchContent: generated protos for Windows")
+
+FetchContent_Declare(
+    protos
+    URL https://github.com/jlknauer/cpp-sc2/releases/download/v1.4.1/generated-s2clientprotocol-db14236-protobuf-1a74ba4.zip
```
alkurbatov, line 5: Please notify me when we have final version of the bindings for this PR and I'll upload to the S3 store.

CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
 cmake_minimum_required(VERSION 3.15)
 
+# Skip generating proto files with Protobuf compiler and use prebuilt files instead
+# Will still build Protobuf as libprotobuf is required for linking
+option(WSL2_CROSS_COMPILE "Building on WSL2 for Windows" OFF)
```
alkurbatov, line 5: Could be a good idea to force OS check and bail out if target system is not windows or doesn't has installed mingw.

src/sc2protocol/CMakeLists.txt
```diff
@@ -31,21 +32,35 @@ if (MSVC)
...
+            COMMAND
+                "${PROJECT_BINARY_DIR}/bin/protoc"
+                "-I=${sc2protocol_SOURCE_DIR}"
+                "--cpp_out=${PROJECT_BINARY_DIR}/generated"
+                "${proto}"
+            DEPENDS protoc
+            VERBATIM
+        )
+    endforeach()
+endif ()
```
alkurbatov, line 66: Style: Please keep the new lines at the end.

CMakeLists.txt
jlknauer: As we are providing the toolchain file - with the expectant paths for the `mingw` toolchain - configuration will fail in this case:
```
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:10 (project):
  The CMAKE_C_COMPILER:

    /usr/bin/x86_64-w64-mingw3-gcc

  is not a full path to an existing compiler tool.

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.


CMake Error at CMakeLists.txt:10 (project):
  The CMAKE_CXX_COMPILER:

    /usr/bin/x86_64-w64-mingw3-g++

  is not a full path to an existing compiler tool.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
  ```
  Here I've simply provided a bad path for the compilers as an example.
  
  But I do see value in building only if `LINUX`, as this feature has no value when building on Windows alone.

File: CMakeLists.txt [hunk 4]
```diff
@@ -1,5 +1,15 @@
 cmake_minimum_required(VERSION 3.15)
 
+# Skip generating proto files with Protobuf compiler and use prebuilt files
+# Will still build Protobuf as libprotobuf is required for linking
+option(WSL2_CROSS_COMPILE "Building on WSL2 for Windows" OFF)
+if (WSL2_CROSS_COMPILE AND UNIX)
```
alkurbatov, line 6: This condition a bit wrong as UNIX includes OS X.
Should be e.g.

```cmake
if (WSL2_CROSS_COMPILE)
    if (UNIX AND NOT APPLE)
        set(CMAKE_TOOLCHAIN_FILE cmake/toolchain/x86-64-w64-mingw32.cmake)
    else ()
        message(FATAL_ERROR "Invalid environment.\
                        WSL2_CROSS_COMPILE=ON for WSL only")
    endif ()
endif ()
```

src/sc2api/sc2_connection.cc
```diff
@@ -8,7 +8,9 @@
 
 #include "civetweb.h"
 
-#if defined (_WIN32)
+#if defined (_WIN32) && defined(__MINGW32__)
+#include <winsock2.h>
+#elif defined (_WIN32) && !defined(__MINGW32__)
```
alkurbatov, line 13: I think this condition could be shortened to:
`#elif defined (_WIN32) `

alkurbatov (commented):
Looks good to me, except several issues. The issues are minor and I can accept the current patch. Lets me know if you are going to fix them.

jlknauer:
Recommended changes have been made and rebased in.

CMakeLists.txt
```diff
@@ -1,5 +1,15 @@
 cmake_minimum_required(VERSION 3.15)
 
+# Skip generating proto files with Protobuf compiler and use prebuilt files
+# Will still build Protobuf as libprotobuf is required for linking
+option(WSL2_CROSS_COMPILE "Building on WSL2 for Windows" OFF)
+if (WSL2_CROSS_COMPILE AND UNIX AND NOT APPLE)
+    set(CMAKE_TOOLCHAIN_FILE cmake/toolchain/x86-64-w64-mingw32.cmake)
+elseif (WSL2_CROSS_COMPILE AND WIN32)
+    message(FATAL_ERROR "Invalid environment.\
```
alkurbatov, line 9: Well, actually it'll not throw if we have APPLE and WSL2 on. Thats why I used more complex condition before.

alkurbatov (approved):
Thanks!
Looks good.

PR #121, opened 2022-08-26
build: wsl2 downstream project support

jlknauer:
Downstream projects must provide a toolchain file, which will set and
unset CMake OS variables which would otherwise trigger the FATAL_ERROR
in cpp-sc2 for detecting invalid environments.


alkurbatov (approved):
Thank you!
Also, extra documentation looks quite helpful.

PR #124, opened 2022-08-31
docs: minor readme changes

jlknauer:
ipv6-parse was removed as a dependency in
51c963cb5668bc48a92dbe4cd9b0ff512bf7449c.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ cpp-sc2
 [![Build Status](https://github.com/cpp-sc2/cpp-sc2/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/cpp-sc2/cpp-sc2/actions/workflows/ci.yml)
 
 The StarCraft II API provides access to in-game state observation and unit control. The API is a wrapper around protobuf defined protocol over a websocket connection.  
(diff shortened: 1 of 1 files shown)


jlknauer:
A small documentation PR, but also raising the question regarding the [precompiled libraries section](https://github.com/cpp-sc2/cpp-sc2#precompiled-libraries) in the README, as these are quite old (~2 years), and the current build configuration does not support linking against precompiled libraries out of the box.

*This is* something I could easily integrate as a `FetchContent` dependency - similar to how we handled the precompiled protos. I do agree with your original sentiment that there should be CI which handles the building, packaging, and uploading of these files to your storage. Whether this is something you'd be interested in having integrated into `cpp-sc2` is up to you - I'm fine with either outcome since the compile overhead is minor and the project really does build right out of the box.

alkurbatov:
Hi. Thank you for the corrections.

The precompiled libraries are for people who know they need what to do with them as there were several requests to get compiled libraries in the past. Also, some older projects like command center still rely on some precompiled artifacts (might be broken anyways). In blank bot I prefer the current approach with submodules/fetch from repo mostly because it is simpler to develop.

I'am going to setup libraries publishing routine, but I don't think that we need to prefer this approach over the current.

PR #128, opened 2022-09-22
chore add .clang-format using "clang-format -style=google -dump-confi…

### LINKED ISSUE #52: add .clang-format file
I realize the guideline says to use Google style formatting except for 4 space tabs, but this isn't actually the case. For example, Lines are not limited to 80 chars, include files are not always sorted, "private/public:" are at times not indented as in Google, etc. Does it make sense to put a .clang-format file in the source tree Visual Studio document formater can be used without fear of munging the code? 

```
BasedOnStyle:  Google
IndentWidth:   4
ColumnLimit: 0
# AlignConsecutiveAssignments: true  # I like this, but I'm not sure you'd agree. It sometimes requires an extra blank line to make it pretty
# AlignConsecutiveDeclarations: true  # as above

SortIncludes: false
...
```

Nickrader:
…g > .clang-format" from WSL2
Change indent width to 4 space.
Resolves #52 
However, current style from Blizzard seem doesn't use 80 for Column Limit. 
Which is a noticeable difference on a file like sc2_agent.cc with it's long declarations.

EditorConfig doesn't seem to support column limits on MS Visual Studio, 
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#supported-by-a-limited-number-of-editors

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,168 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
+AccessModifierOffset: -1
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: Never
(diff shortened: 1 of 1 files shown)


alkurbatov:
Thank you! We indeed planned to do code reformatting at some point, but right now I would prefer to postpone this change.
There are many reason discussed in the discord channel. I think we should switch to autoformatting slowly.

Right now the nearest step I see is to apply formatting tools to the cmake files as the style is more or less consistent there.

Nickrader:
Cool.  I suspected it might be a more involved task, with a lot of diffs to sort through.  As even when I put Column Limit at 120 or above, it then changes many things to one-liners.  No magic bullet in either extreme.

PR #131, opened 2023-10-07
feat: Update dependencies and support VS 2022

alkurbatov:


alkurbatov:
@Nickrader this might be enough to support new Visual Studio on the API's side.  
Could you please check if you have some time to spare?

Nickrader:
I was unable to connect to game when I tried to run all_tests.exe or an example.

alkurbatov:
> I was unable to connect to game when I tried to run all_tests.exe or an example. I would time out at "Waiting for connection". I'll try it again tomorrow.

That my be an issue with the game itself as sometimes it tries to download some data or contact battle net. This could be verified by manual starting the game without any bots.

Nickrader:
Idk, they seem to have done something in Civetweb over the last year, that changed life for windows users, is my best guess? 

If I link this branch to an old civetweb, https://github.com/Nickrader/civetweb_old/commits/master , (salvaged from a build of command center), tests seem to run fine (except TestTrainBainling, which I think was existing issue, the fragile Blizztests) .  
From what I can tell, you put cpp-sc2 changes on top of where they always went for the new civetweb.  

I am unsure how I would go about narrowing the cause?  It seems regression testing civetweb would be a giant PITA since we place our changes on-top?  I'm open to suggestions.

I did notice when cmake was building project, my civetweb did state targeting: sc2 5.0.10 in the terminal... However, when ran all_tests.exe, it targeted newest sc2; Base91115

alkurbatov:
>From what I can tell, you put cpp-sc2 changes on top of where they last went for the new civetweb.

This is usual rebase approach I do from time to time. Shouldn't be a problem, although I am not sure we still need these patches.

>I am unsure how I would go about narrowing the cause? It seems regression testing civetweb would be cumbersome since we place our changes on-top? I'm open to suggestions.

Could be hard to tell, probably some debug logs or even debugger could help here, but I believe I should revert civetweb back. It seems that we can compile with older civetweb on Windows and new VC. The new connection issue will be investigated in https://github.com/cpp-sc2/cpp-sc2/issues/133

I'll revert the changes in civetweb.

alkurbatov:
As for the flaky test: I think something has changed in type ids with one of the latest patches and this is real bug. At least we need to update type id mappings.

alkurbatov:
@Nickrader seems that civetweb now uses IPv6 by default and it could be the reason of connection timeout. Could you please check the new version with disabled IPv6?

// Seems that I should finally buy some windows hardware simply to test changes :(

Nickrader:
That didn't fix it for me.  I cannot rule out it being a 'me' problem, since I have little understanding of networking (and I've done some things like blocking Blizz in Australia and SEA in my firewall).  We could always solicit help on discord, some people there seem to be running Windows also.  
I only have Windows hardware because I never knew any better ;)  Originally got me locked-in with consoles, and then PC games.  

I think your proposal to revert Civetweb may be the best.  

`// Seems that I should finally buy some windows hardware simply to test changes :( `
I guess that depends on how scarce money is to you ;)  I wouldn't, lol.  Next PC I build will probably be Linux only, unless they make Windows free.

alkurbatov:
> I guess that depends on how scarce money is to you ;) I wouldn't, lol. Next PC I build will probably be Linux only, unless they make Windows free.

I would prefer Linux or OS X, but the game client is quite old and there is almost no chance for updates. However, most of our users use Windows and many of them not professional programmers. So I simply tired to do remote debugging using chat :)
Another problem here: many fixes require careful testing and debugging on Windows, which is almost blocker for me.

> I think your proposal to revert Civetweb may be the best.

Agree, will do this evening.

alkurbatov:
Reverted.

Nickrader (approved):
all_tests.exe ran fine.  Lgtm.

PR #137, opened 2023-12-09
feat: Update CMakeLists for Visual Studio

Nickrader:
Added include sources to cmake, needed for "header" and "source" folders to show up in Visual Studio Solution Explorer.
Also added debug arguments, so one can simply hit "Windows Debugger" in the IDE, to get the tests to run.  
Hopefully will make troubleshooting Windows installs easier, by not having to manually set these defaults.

Open to any changes/suggestions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,9 @@ if (MSVC)
 
     # Use statically linked runtime
     set(CMAKE_MSVC_RUNTIME_LIBRARY MultiThreaded$<$<CONFIG:Debug>:Debug>)
+
+    # Set startup project in Visual Studio
+    set_property( DIRECTORY PROPERTY VS_STARTUP_PROJECT all_tests)
 endif ()
 
 if (BUILD_API_EXAMPLES)
--- a/src/sc2api/CMakeLists.txt
+++ b/src/sc2api/CMakeLists.txt
@@ -26,7 +26,35 @@ set(sc2api_sources
     sc2_unit.cc
     "typeids/sc2_${SC2_VERSION}_typeenums.cpp")
 
-add_library(sc2api STATIC ${sc2api_sources})
+set(header_path "${PROJECT_SOURCE_DIR}/include/sc2api/")
+set(sc2api_headers
+    ${header_path}/sc2_action.h
+    ${header_path}/sc2_agent.h
+    ${header_path}/sc2_api.h
+    ${header_path}/sc2_args.h
+    ${header_path}/sc2_client.h
+    ${header_path}/sc2_common.h
+    ${header_path}/sc2_connection.h
+    ${header_path}/sc2_control_interfaces.h
+    ${header_path}/sc2_coordinator.h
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -66,6 +66,9 @@ if (MSVC)
 
     # Use statically linked runtime
     set(CMAKE_MSVC_RUNTIME_LIBRARY MultiThreaded$<$<CONFIG:Debug>:Debug>)
+
+    # Set startup project in Visual Studio
+    set_property( DIRECTORY PROPERTY VS_STARTUP_PROJECT all_tests)
```
alkurbatov, line 71: Could you please tell why we should start from all_tests and not e.g. sc2lib?

Also, this string contains extra space after the opening `(`.

src/sc2api/CMakeLists.txt
```diff
@@ -26,7 +26,35 @@ set(sc2api_sources
     sc2_unit.cc
     "typeids/sc2_${SC2_VERSION}_typeenums.cpp")
 
-add_library(sc2api STATIC ${sc2api_sources})
+set(header_path "${PROJECT_SOURCE_DIR}/include/sc2api/")
+set(sc2api_headers
+    ${header_path}/sc2_action.h
```
alkurbatov, line 31: I think you should apply list transform to properly modify all paths and reduce repetition.
See https://cmake.org/cmake/help/latest/command/list.html#transform

alkurbatov (commented):
Hey @Nickrader 
Thank you for the patch and excuse me for long review.

The idea looks good, but the realization could be improved a bit.

alkurbatov:
> Hey @Nickrader Thank you for the patch and excuse me for long review.
> 
> The idea looks good, but the realization could be improved a bit.

Also, this PR should be marked as refactor or build as we do not andd any new features from the API's (product) point of view.

src/sc2api/CMakeLists.txt
alkurbatov: Not needed after the recent changes.

alkurbatov (changes requested):
@Nickrader 
Please, check on the latest version of the project. Should be good now.

CMakeLists.txt
Nickrader: The logic was after downloading and building the project, the next step was to run all_tests.

It was an attempt to 'simplify' the process of getting something to show up as a newbie using Visual Studio.  
All the 'VS_****_****' commands were to facilitate using the IDE buttons, once entering Visual Studio, and not using the shell. 
However, one using Visual Studio will eventually have to learn to manually modify these values in the IDE, so perhaps these initial quality-of-life modifications aren't needed.  I was on a bit of an 'automation' kick that day, learning more about CMAKE.

Nickrader:
Headers now show up in Visual Studio IDE.  
The other 'VS_****_****' commands are optional, not sure if others find it useful for the first-time setup or if explanations/pictures how to set these variables in IDE is more valuable.

PR #139, opened 2023-12-30
Unify sources

alkurbatov:


alkurbatov:
Hey @Nickrader 
This patch fixes some the issues with headers location in VS. This also looks convenient to keep all files together, especially because we do not differ headers into public and private.

alkurbatov:
Seems that it works well.

PR #142, opened 2024-04-19
docs: Update linux build instructions

### LINKED ISSUE #141: Building Linux parallel parameter shell freezes
Documentation, https://github.com/cpp-sc2/cpp-sc2/blob/master/docs/building.md#linux-cmdline ,  shows building using parameter `--parallel`.

This can cause shell to lockup when building protobuf.
`Building CXX object _deps/protobuf-build/CMakeFiles/libprotobuf.dir/src/google/protobuf/compiler/subprocess.cc.o`

Workaround to pass 'n-1' threads your cpu supports to `--parallel` parameter.
E.g. `cmake --build build --parallel -j 7`  assuming your cpu supports 8 threads.

**To Reproduce**
Follow build steps using Linux, leaving '--parallel' as is.

**Expected behavior**
Finish compilation of project

**Desktop (please complete the following information):**
 - OS / Distro:  Ubuntu 23.10

I am assuming this would fall under documentation change?  Or maybe fails elsewhere for others?

Nickrader:
Building protobuf with '--parallel' parameter causes computer to freeze and shell to crash. 
Using 'n-1' threads that your cpu supports seems to avoid google crashing the shell.

Tried building protobuf outside the project with '--parallel' and it broke similar place as in our project.

closes #141

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -76,7 +76,9 @@ $ cd cpp-sc2
 $ cmake -B build
 
 # Build the project.
-$ cmake --build build --parallel
+$ cmake --build build --parallel $(nproc --ignore=0)
+
+# If building freezes in above step, increase number: --ignore=1, ignore=2 ...
 ```
 
 ## Linux (cmdline)
@@ -91,7 +93,9 @@ $ cd cpp-sc2
 $ cmake -B build
 
 # Build the project.
-$ cmake --build build --parallel
+$ cmake --build build --parallel $(nproc --ignore=0)
+
+# If building freezes in above step, increase number: --ignore=1, ignore=2 ...
 ```
 
 ## Compilation options


docs/building.md
```diff
@@ -91,7 +91,9 @@ $ cd cpp-sc2
 $ cmake -B build
 
 # Build the project.
-$ cmake --build build --parallel
+# Note: number after '-j' is threads used.
+# Max is 'n-1' threads your cpu supports.
+$ cmake --build build --parallel -j 4
```
alkurbatov, line 96: Do we have this issue only in Linux? At least macOs has the same problem.

Also, we can try to use all available core - 1 e.g. by doing:
https://unix.stackexchange.com/questions/208568/how-to-determine-the-maximum-number-to-pass-to-make-j-option

alkurbatov: The following works on mac:

```bash
$ nproc
12

$ nproc --ignore=1
11
```

alkurbatov: > From what I read, it's not just about jobs, but the memory it uses. So when google uses 9 jobs, it is over 13.5 GBs on my machine?)

Ah, that could be. Nice catch!

Ok, lets do the following:
1. For macOs and Linux add `$(nproc)`.
2. For windows put `4`, but add comment that one can increase this value according to his cpu count.

Nickrader:
I tried including the `ignore` parameter.  I didn't encounter any problems using `nproc`, so I left it at zero, with a note to increase if building freezes.  
Didn't have any problems building on windows command line.  Probably have to use `make for windows` https://gnuwin32.sourceforge.net/packages/make.htm as build system to encounter same behavior.

docs/building.md
```diff
@@ -76,7 +76,9 @@ $ cd cpp-sc2
 $ cmake -B build
 
 # Build the project.
-$ cmake --build build --parallel
+$ cmake --build build --parallel $(nproc --ignore=0)
+
+# If building freezes in above step, increase number: --ignore=1, ignore=2 ...
```
alkurbatov, line 81: Lets reshape it in the following way:

```markdown
# Build the project.
cmake --build build --parallel $(nproc)

# If building freezes in above step, decrease nproc to 1 or more, e.g.:
cmake --build build --parallel $(nproc --ignore=1)
```

alkurbatov: Since the first command must work for everyone, we provide fallback command to those who have build issues.

docs/building.md
alkurbatov: Addressed in https://github.com/cpp-sc2/cpp-sc2/commit/e4e7bafca81d5266da811370bf44503729a0fd41 with suggested corrections.

Thank you for the patch!

Issue #143, opened 2024-05-05
Unable to build for ladder using wsl2


nanogeo:
**Describe the bug**
Build fails with errors saying "This file was generated by an older version of protoc which is incompatible with your Protocol Buffer headers. Please regenerate this file with a newer version of protoc."

**To Reproduce**
Steps to reproduce the behavior, for example:
1. Download fresh install of Ubuntu with wsl2
2. Run the following commands:
    1. apt-get update
    2. sudo apt-get install cmake
    3. sudo apt install build-essential
    4. sudo apt install mingw-w64-x86-64-dev  g++-mingw-w64-x86-64  gcc-mingw-w64-x86-64 win-iconv-mingw-w64-dev
    5. sudo update-alternatives --config x86_64-w64-mingw32-gcc
       - Choose option 1
    6. sudo update-alternatives --config x86_64-w64-mingw32-g++
       - Choose option 1
3. Clone blank_bot repo (https://github.com/cpp-sc2/blank-bot)
4. Navigate to said repo
5. Generate CMake configuration with "cmake -B build -DWSL2_CROSS_COMPILE=ON"
6. Compile project with "cmake --build build -j"
7. Observe numerous errors similar to:
```
/mnt/c/Users/hocke/OneDrive/Documents/blank_bot/build/_deps/cpp_sc2-build/generated/s2clientprotocol/debug.pb.h:17:2: error: #error This file was generated by an older version of protoc which is
   17 | #error This file was generated by an older version of protoc which is
      |  ^~~~~
/mnt/c/Users/hocke/OneDrive/Documents/blank_bot/build/_deps/cpp_sc2-build/generated/s2clientprotocol/debug.pb.h:18:2: error: #error incompatible with your Protocol Buffer headers. Please
   18 | #error incompatible with your Protocol Buffer headers. Please
      |  ^~~~~
/mnt/c/Users/hocke/OneDrive/Documents/blank_bot/build/_deps/cpp_sc2-build/generated/s2clientprotocol/debug.pb.h:19:2: error: #error regenerate this file with a newer version of protoc.
   19 | #error regenerate this file with a newer version of protoc.
      |  ^~~~~
```

**Expected behavior**
The build should finish without errors.

**Desktop (multiple version means multiple attempts with different version):**
 - OS / Distros: Ubuntu 20.04.6/22.04.3
 - cpp-sc2 Version: 1.4.1
 - StarCraft II version 5.0.12
[...]

PR #145, opened 2024-09-19
fix: Update WSL2 proto assets, #143

alkurbatov:
To fix WSL2 build with the current protobuf.


alkurbatov:
@Nickrader please double check just in case.

Nickrader:
Windows executable files are built.  No errors reported.

alkurbatov:
Thank you! Seems that we finally fixed it :)

Issue #133, opened 2023-10-10
Investigate the reason of connection timeout with new civetweb


alkurbatov:
After update to civetweb v1.16 we experience connection timeout on Windows and game could not be started. Works well at least on OS X.

We need to investigate the reason as it possibly blocks update of the submodule.

Nickrader:
Civetweb:
Results of regression test.  
Good: c1cf4be2d3a1c118c64a815672e215af0772ffbe
Bad: b33c879584e0fce2af0b0be1f6e472e7f7be7ab1

Seems Master branch commit history is changed between some commits in this range.  Rebased?

I was unable to figure out how to get `error_log_file` to output anything or create a file?
https://github.com/civetweb/civetweb/blob/master/docs/UserManual.md#error_log_file

Tried adding option to `sc2_connection.cc` ;  function `StartCivetweb()` 
```	
static const char* LOG_ERROR = "error.log";
	
options[] = {    
    ....
    "error_log_file",
    LOG_ERROR,
    ....
}
```
That did not seem to do anything.

```
https://github.com/civetweb/civetweb/blob/master/docs/UserManual.md
	Embedding with OpenSSL on Windows might fail because of calling convention. 
	To force CivetWeb to use __stdcall convention, add /Gz compilation flag in Visual Studio compiler.
```
Changing calling convention did not fix.

Common point of divergence seems to happens sometime after `setsockopt` gets called in `civetweb.c` , however this enters loop shortly after and in good commit, after 4-8? passes, connects to game.  Bad commit times out.  I get lost here.
```
    good commit: pollres == 1 (eventually)
    bad: pollres != 1 (ever)
```

The answer may lie in some error logs, I'll continue to play around trying to access, but this may be as far as I can follow at the moment.  A lot of stuff happens in `civetweb.c` that I don't follow at the moment, my first time looking into a web server.

alkurbatov:
>Seems Master branch commit history is changed between some commits in this range. Rebased?

If by master you mean my fork of civetweb, it was rebased with changes in history (bad approach although simple).

I would suggest to try the latest version of civetweb. It could contains fixes of the problem.

Another option here is to try civetweb without these 2 patches from Blizzard as they could brake something.

Nickrader:
I forked directly from Civetweb then rebased Blizz changes on that. 

[...]

PR #154, opened 2025-10-06
Patch civetweb cmake min version for cmake4

### LINKED ISSUE #153: cpp-sc2 doesn't build on CMake version 4
**Describe the bug**
cpp-sc2 doesn't build on CMake version 4, the build fails in the cmake configure step

**To Reproduce**
Steps to reproduce the behavior, for example:
1. Install Cmake 4
2. Try to build the repo, or use it from another build
3. Observe `Compatibility with CMake < 3.5 has been removed from CMake.`

**Expected behavior**
It should build.

**Desktop (please complete the following information):**
 - OS / Distro: Arch Linux
 - cpp-sc2 Version: master
 - StarCraft II version: N/A

**Additional context**
Reports of this in the ai arena discord as well. This appears to be caused by the pinned civetweb version (1.15) using an ancient cmake version.

ccapitalK:
Fixes #153

### DIFF
--- a/thirdparty/civetweb/0003-Patch-CMake-Min-Version.patch
+++ b/thirdparty/civetweb/0003-Patch-CMake-Min-Version.patch
@@ -1,0 +1,25 @@
+From 1b3b1f65d6fb872d66d1d6d49b346d404a7d40bc Mon Sep 17 00:00:00 2001
+From: Sahan Fernando <sahan.h.fernando@gmail.com>
+Date: Mon, 6 Oct 2025 21:30:23 +1100
+Subject: [PATCH] Patch CMakeLists
+
+---
+ CMakeLists.txt | 6 ++----
+ 1 file changed, 2 insertions(+), 4 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 1eb391f2..041b2295 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1,7 +1,5 @@
+-# Use at least CMake 3.3
+-cmake_minimum_required (VERSION 3.3.0)
+-cmake_policy(VERSION 3.2.2)
+-cmake_policy(SET CMP0054 NEW)
++cmake_minimum_required (VERSION 3.10)
++cmake_policy(VERSION 3.10)
+
+ # Set up the project
+ project (civetweb)
+--
+2.51.0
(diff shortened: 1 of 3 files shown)


ccapitalK:
Oops, put an unrelated commit into here, will fix this pr to not have the second one.

Edit: Fixed

ccapitalK:
So, I fixed the lint failure (removed the trailing empty line in the patch file). Not sure what the best way to fix the osx abseil failure is, I'm not 100% certain that it's related to this change (abseil is pulled in by protobuf, not civetweb, right?) I do have a second fix to bump the abseil version as well, on https://github.com/ccapitalK/cpp-sc2/tree/ccapitalK-modern-build, would it make sense to put up a PR for both together?

> clang++: error: unsupported option '-msse4.1' for target 'arm64-apple-darwin24.6.0'

alkurbatov:
Hi @ccapitalK 
I would suggest to update CivetWeb dependency instead as they finally upgraded cmake, [see](https://github.com/civetweb/civetweb/blob/b6ef58f4c4c7fbe90fd1065bccf45b143345f1a6/CMakeLists.txt#L1). This could cause some weird issues on windows however, but we haven't checked them on the recent update.

alkurbatov:
As for abseil, it seems that we have to update protobuf dependency again. This could be down in this PR.

ccapitalK:
> I would suggest to update CivetWeb dependency instead as they finally upgraded cmake, [see](https://github.com/civetweb/civetweb/blob/b6ef58f4c4c7fbe90fd1065bccf45b143345f1a6/CMakeLists.txt#L1). This could cause some weird issues on windows however, but we haven't checked them on the recent update.

I looked through the history of the repo. Civetweb was downgraded to 1.15 in #150, pending #133 being investigated, right? Was there any progress made on that issue?

ccapitalK:
Adding second commit to patch upgrade protobuf.

alkurbatov:
> * CI / build (windows-latest) (pull_request)

Unfortunately no progress. My guess is that Civetweb has improved from the last time we experienced these issues.

thirdparty/civetweb/0003-Patch-CMake-Min-Version.patch
```diff
@@ -0,0 +1,25 @@
...
+index 1eb391f2..041b2295 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1,7 +1,5 @@
+-# Use at least CMake 3.3
+-cmake_minimum_required (VERSION 3.3.0)
+-cmake_policy(VERSION 3.2.2)
+-cmake_policy(SET CMP0054 NEW)
++cmake_minimum_required (VERSION 3.10)
++cmake_policy(VERSION 3.10)
```
alkurbatov, line 20: Seems that something is wrong with the patch, CI failed:

https://github.com/cpp-sc2/cpp-sc2/actions/runs/18444774019/job/52549914866?pr=154

alkurbatov:
Seems the issue with Windows looks weird and hard to debug I created a separate branch to keep the latest changes. The branch has updated Civetweb and Protobuf, and some minor improvements.
@ccapitalK please switch to v2.1.0 and thank you for help.

PR #155, opened 2026-05-05
Fix 3 proto : pod conversion bugs + bump sc2protocol pin

jalcim:
## Bugs fixed

- `ScoreDetails::collected_vespene` was never set in proto→pod (always 0). `sc2_proto_to_pods.cc`
- `total_damage_taken` gated by `has_collected_minerals()` instead of `has_total_damage_taken()`, value lost when minerals field absent. `sc2_proto\
_to_pods.cc:100`
- `AbilityData::is_instant_placement` assigned `has_is_instant_placement()` (presence) instead of the actual value, always true when field present.\
 `sc2_data.cc:125`

## Infra

- Bump sc2protocol pin `c04df4a` → `bff45da` (v5.0.12 → v5.0.15). Byte-identical .proto + stableid.json, only version string differs.
- Bump civetweb v1.15 → `dfdfdb6` and protobuf v3.23.4 → v33.0 for CMake v4 compatibility (mirrors upstream commit `349f451` from v +2.1.0, #153).

### DIFF
--- a/src/sc2api/sc2_data.cc
+++ b/src/sc2api/sc2_data.cc
@@ -122,7 +122,7 @@ void AbilityData::ReadFromProto(const SC2APIProtocol::AbilityData& ability_data)
 
     is_instant_placement = false;
     if (ability_data.has_is_instant_placement()) {
-        is_instant_placement = ability_data.has_is_instant_placement();
+        is_instant_placement = ability_data.is_instant_placement();
     }
 
     cast_range = 0.0f;
--- a/src/sc2api/sc2_proto_to_pods.cc
+++ b/src/sc2api/sc2_proto_to_pods.cc
@@ -97,7 +97,7 @@ bool Convert(const ScorePtr& score_ptr, ScoreDetails& score_details) {
         Convert(score_details_ptr->total_damage_dealt(), score_details.total_damage_dealt);
     }
 
-    if (score_details_ptr->has_collected_minerals()) {
+    if (score_details_ptr->has_total_damage_taken()) {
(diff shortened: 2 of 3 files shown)


PATCHES.md
```diff
@@ -0,0 +1,85 @@
+# Patches divergents kasper-side
```
alkurbatov: Hi @jalcim 
Thank you for the fix!

Could you please tell me why do we need PATCHES.md?

jalcim: Just forget to remove ^^

alkurbatov: Ok.
Please, remove it and I'll accept the patch. The CI failure is definitely because we need some changes in the build routine.

jalcim:
I added `Bump civetweb v1.15 → dfdfdb6 and protobuf v3.23.4 → v33.0 for CMake v4 compatibility (mirrors upstream commit 349f451 from v +2.1.0, https://github.com/cpp-sc2/cpp-sc2/issues/153).`

I dont know how you wanna managing that, but CI was crying.

alkurbatov:
This was actually fixed in https://github.com/cpp-sc2/cpp-sc2/tree/v2.1.0
I pushed the changed to master, please rebase. Should work now.

jalcim:
> This was actually fixed in https://github.com/cpp-sc2/cpp-sc2/tree/v2.1.0 I pushed the changed to master, please rebase. Should work now.

done

alkurbatov (approved):

alkurbatov:
Thank you! LGTM

Issue #71, opened 2020-05-11
Setup unit tests routine


alkurbatov:
We have lots of code which could be tested in some way, e.g. the new filters. Unfortunately, the only tests we have are functional requiring launch of a game client.

This task is about setting up a number of small unit test suites to check at least the filters.

alkurbatov:
Well, the simplest way possible is to add GoogleTest as submodule and drop all that hand-maderoutine the Blizzard devs wrote. While this will be not real unit-tests (we need a running client for run), we can add basic unit-tests, too.

Nickrader:
https://github.com/Nickrader/cpp-sc2/tree/dev_unittest

I started off trying to implement a unit test routine.  I have since moved the goalpost to completing a sample, covering the different test types in the GTest framework.  [it's taking me a long time to translate between Google documentation and our API, since it's my first go at GTest and real-world code].

Think I'm far enough to open it up for feedback.  Figured getting a sample done first would make it more accessible for future contributors.  

Currently still a work-in-progress.

alkurbatov:
First of all, this is definitely a good attempt and nice way to get useful experience.
Regarding the code the original idea is the following:
1. Link with googletest (seems that we should use FetchContent here as I saw quite a lot of problems with "simple install googletest").
2. Use the same flag (BUILD_API_TEST) to build all kinds of tests.
3. Replace custom test routine in the integration tests with google test things.
4. Add some unit tests (e.g. for unit filters) just to test that it works.
5. Add unit-tests run into github CI.

I am totally ok if you wish to implement some points from this list as whole list could take a while.

alkurbatov:
By the way, perhaps you may find Catch2 more convenient to use in tests:
https://github.com/catchorg/Catch2/blob/devel/docs/tutorial.md

PR #156, opened 2026-05-05
Bundled fixes, missing proto-field exposures, and doctest unit tests

### LINKED ISSUE #62: Prevent crash when observation is missing
I sometimes (but very rarely) get a crashed caused by a missing observation. This is what is outputed in stdout by the API.

```Error in returning observation:
The main response is of type: 12
There is no ResponseObservation/message!
No error strings in result.

Maybe the error can still be outputed but not make the bot crash and just wait for the next frame?

### PR DESCRIPTION
**Pull request by jalcim**
10 patches accumulated,
three bugs and seven additions.

- Connection.Receive could leak a Response and race the queue when a timeout fired
- The converter would abort the entire tick if SC2 returned an out-of-range enum (happens with mods or custom maps)
- Coordinator used to call exit(1) when SC2 failed to start, killing the process before the client could clean up. Now throws ClientConnectionError instead
- UnitOrder.target_pos was truncated to 2D, terrain altitude lost. Now Point3D (still backward-compatible since Point3D exposes .x and .y)
- Effect.alliance / owner / radius: useful for Scanner Sweep (knowing if it's the opponent scanning you), Blinding Cloud or Corrosive Bile in team games (don't blanket your allies), and identifying which opponent landed an EMP in 2v2 / 3v3 via owner
- 4 Unit fields never extracted: is_active, buff_duration_remain / max, rally_targets
- RadarRings: your own Sensor Towers and Ravens in detect mode, never read before
- action_errors: the red "Not enough minerals" / target invalid / etc. errors from the UI are now exposed, so client code can react the same frame instead of waiting for a timeout
- 20 alerts that were silently dropped now route to dedicated OnAlert\* hooks. Before, only NuclearLaunchDetected and NydusWormDetected made it through. The other 20 (BuildingUnderAttack, MULEExpired, TrainError, etc.) were ignored
- Doctest-based unit test framework, with two regression tests on already-merged fixes (collected_vespene, is_instant_placement). Lets you write tests without launching the SC2 client

Closes #62, #71, #149.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ set(FETCHCONTENT_QUIET FALSE)
 option(BUILD_SC2_RENDERER "Build SC2 Renderer library" ON)
 option(BUILD_API_EXAMPLES "Build Examples" ON)
 option(BUILD_API_TESTS "Build Tests" ON)
+option(BUILD_UNIT_TESTS "Build doctest-based unit tests (no game client required)" OFF)
 
 set(SC2_VERSION "5.0.14" CACHE STRING "Version of the target StarCraft II client")
 message(STATUS "Target SC2 version: ${SC2_VERSION}")
@@ -76,6 +77,10 @@ endif ()
 
 if (BUILD_API_TESTS)
     add_subdirectory(tests)
+endif ()
+
+if (BUILD_UNIT_TESTS)
+    add_subdirectory(tests/unit)
 endif ()
 
 if (WSL2_CROSS_COMPILE)
--- a/src/sc2api/sc2_client.h
+++ b/src/sc2api/sc2_client.h
@@ -120,44 +120,64 @@ public:
 
     //! \name Extended alert hooks. All default to noop, override only what you use.
     //!@{
+    //! Called when a generic in-game error alert is raised.
     virtual void OnAlertError() {
     }
+    //! Called when an add-on (Reactor / Tech Lab) finishes construction.
     virtual void OnAlertAddOnComplete() {
     }
+    //! Called when a building finishes construction.
     virtual void OnAlertBuildingComplete() {
     }
+    //! Called when one of the player's buildings is under attack.
     virtual void OnAlertBuildingUnderAttack() {
     }
+    //! Called when a larva hatches.
     virtual void OnAlertLarvaHatched() {
     }
+    //! Called when an Archon merge completes.
     virtual void OnAlertMergeComplete() {
     }
+    //! Called when a mineral patch is exhausted.
     virtual void OnAlertMineralsExhausted() {
     }
+    //! Called when a unit morph completes.
     virtual void OnAlertMorphComplete() {
     }
+    //! Called when a Mothership finishes construction.
     virtual void OnAlertMothershipComplete() {
     }
+    //! Called when a MULE expires.
     virtual void OnAlertMULEExpired() {
     }
+    //! Called when a Nuke launch completes.
     virtual void OnAlertNukeComplete() {
     }
+    //! Called when a research alert is raised.
     virtual void OnAlertResearchComplete() {
     }
+    //! Called when a training order fails.
     virtual void OnAlertTrainError() {
     }
+    //! Called when a unit finishes training.
     virtual void OnAlertTrainUnitComplete() {
     }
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

**Comment by jalcim** [comment]
I wait the review for send all the test

File: CMakeLists.txt [hunk 1]
```diff
@@ -78,6 +79,10 @@ if (BUILD_API_TESTS)
     add_subdirectory(tests)
 endif ()
 
+if (BUILD_UNIT_TESTS)
```
alkurbatov, line 82: For such cases you should expand one of the existing API tests as most of the fixes affect interactions with the proto part.

src/sc2api/sc2_client.h
```diff
@@ -118,6 +118,50 @@ class ClientEvents {
     virtual void OnNuclearLaunchDetected() {
     }
 
+    //! \name Extended alert hooks. All default to noop, override only what you use.
```
alkurbatov, line 121: It would be helpful to document what each call does as it was done for other calls above.

CMakeLists.txt
jalcim: This PR closes #71, which explicitly asks for unit tests that don't require launching the game client.
Coverage for the fixes in this PR is on the follow-up branch `pr/api-coverage-and-fixes-tests`.

jalcim: Most fixes here are pure POD/proto plumbing; a real-game test would add runtime cost without added coverage.

alkurbatov: Well, the problem with this PR is that it goes against the current testing approach of the project.
1. We don't have coverage and thus don't care about it.
2. We already have many tests that runs game client and test the API.
3. It introduces some test framework without any conversation in a way which goes against the project structure.

Sorry but I cannot accept this PR.

jalcim:
Quick follow-up on the framework choice:
the PR uses doctest, not GoogleTest.
This is a simplification, not a deviation.
doctest is single-header and FetchContent-friendly, no library to build or embed,
in the same family as Catch2 which you suggested in #71 on 2023-11-18.
For the surface tested here (proto-to-POD copies, small Convert functions, struct defaults)
GoogleTest would add CMake and build complexity without coverage benefit

jalcim:
So ?

alkurbatov:
Sorry for the late answer, was busy with other tasks.

Regarding the testing framework:
I don't think that compilation speed is the main factor in selecting proper technology. Rather I would prefer take a look at features comparison. This question needs deeper research from my side, can't say go or no-go right now.

Regarding necessity to test against real client:
The problem with all this proto-related code is that Blizzard doesn't follow their own spec well. Yes, they declare that some feature should be reported this way or another, but things could dramatically change when one tries to run it against real client.
Another point: I need a proper way to verify that the suggested functionality actually works and other developers need some examples about it's usage and integration tests solve this problem more or less.
This the quite valuable for proto-related part.

Regarding the PR itself:
I suggest to split it into smaller independent parts (e.g. according to the linked issues) and review separately. Such approach is more clean and clear, and allows me to check the PR part by part.
What do you think about it?

Also, 2vs2, 3vs3 etc never worked. By some reason such modes are not possible in the AI games.

jalcim:
> Sorry for the late answer, was busy with other tasks.
> 
> Regarding the testing framework: I don't think that compilation speed is the main factor in selecting proper technology. Rather I would prefer take a look at features comparison. This question needs deeper research from my side, can't say go or no-go right now.
> 
> Regarding necessity to test against real client: The problem with all this proto-related code is that Blizzard doesn't follow their own spec well. Yes, they declare that some feature should be reported this way or another, but things could dramatically change when one tries to run it against real client. Another point: I need a proper way to verify that the suggested functionality actually works and other developers need some examples about it's usage and integration tests solve this problem more or less. This the quite valuable for proto-related part.
> 
> Regarding the PR itself: I suggest to split it into smaller independent parts (e.g. according to the linked issues) and review separately. Such approach is more clean and clear, and allows me to check the PR part by part. What do you think about it?
> 
> Also, 2vs2, 3vs3 etc never worked. By some reason such modes are not possible in the AI games.

Unfortunately, I no longer have the time.

I proposed something very clean and well-justified.

I've moved forward with my own fork to suit my needs.
It's up to you to see where your analysis leads you.

PR #158, opened 2026-06-01
build: Allow Cross-Compiling Proto files

ShadowLordAlpha:
Adds in respect for the setting Protobuf_PROTOC_EXECUTABLE. If provided the Protoc executable will not be built and we will use the provided file to build the proto files. Do note that it must be the same version as the library so currently that would be v33.x

No changes where made to WSL2_CROSS_COMPILE so if set the library will still use precompiled protos


ShadowLordAlpha:
This was not supposed to push to this yet. Sorry about that

PR #160, opened 2026-08-04
build(cmake): update cmake for ninja; update windows build guide

spaceghost1589:
BLUF: This PR modernizes the CMake build and documentation by introducing a streamlined, automated workflow powered by CMake, Ninja, and CMakePresets.json

Along with a revised build guide (only Windows ATM), this should greatly improve the onboarding experience for new, inexperienced bot developers.


spaceghost1589:
build guide for macOS and Linux included.

spaceghost1589:
Might be smart to pull this into a dev or alpha branch first so people can test the new Ninja/Presets workflow locally, keeping the current setup safe for production ladder builds until it's fully verified.

cmake/toolchain/x86-64-w64-mingw32.cmake
```diff
@@ -1,5 +1,5 @@
 set(CMAKE_SYSTEM_NAME Windows)
-set(CMAKE_SYSTEM_PROCESSOR x86)
+set(CMAKE_SYSTEM_PROCESSOR x86_64)
```
alkurbatov, line 2: question: Are you sure about this? According to the cmake docs, it seems that on Windows x86 os the only valid value. See also https://stackoverflow.com/questions/70475665/what-are-the-possible-values-of-cmake-system-processor

docs/building.md
```diff
@@ -1,130 +1,212 @@
...
-1. Download and install Visual Studio ([2017](https://www.visualstudio.com/downloads/) or newer).
+<details name="prerequisites"><summary>Windows</summary>
 
-1. Clone the project.
+### 1. [Install *Git*](https://git-scm.com/)
 
-   ```bat
-   git clone git@github.com:cpp-sc2/cpp-sc2.git
-   ```
+### 2. Install IDE
```
alkurbatov, line 47: issue: I suggest to remove the IDE step.

1. IDE is something developer should choose manually and before building anything. 
2. Many people don't use IDEs at all and this project can be build we cmake+make+g++.
3. I am totally against promotion of JetBrains IDE since they banned Russian users.

README.md
```diff
@@ -26,39 +27,33 @@ a C++ and class-based abstraction.
...
 [Boosty](https://boosty.to/cpp-sc2). Support of the project allows to make
 releases more often and implement new features for developers.
 
 ## Building the project
 
-This library uses [CMake](https://cmake.org/download/) to generate project
-files. It builds with [Visual Studio](https://www.visualstudio.com/downloads/)
-on Windows and Makefiles on the other platforms. It relies on the following
-contrib packages:
+This library uses [CMake](https://cmake.org/download/) with Ninja to generate project
```
alkurbatov, line 42: question: Could you please describe what is the actual benefit to use ninja here? It adds additional complexity and requires installation of additional tool (at least in Linux).

docs/precompiled_libs.md
```diff
@@ -5,3 +5,9 @@
 
 ## OS X
 [1.3.0](https://storage.yandexcloud.net/cpp-sc2/osx/cpp-sc2-1.3.0.zip) [(changelog)](https://github.com/cpp-sc2/cpp-sc2/releases/tag/v1.3.0)
+
+
+> **DEPRECATED / LEGACY DOCUMENTATION**
```
alkurbatov, line 10: suggestion: Lets remove this file. You're right, it is outdated long time ago.

.clang-format
```diff
@@ -7,3 +7,12 @@ AllowShortIfStatementsOnASingleLine: Never
...
 IndentWidth: 4
+---
+Language: Json
+BasedOnStyle: Google
+AccessModifierOffset: -4
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+ColumnLimit: 120
+IndentWidth: 4
```
alkurbatov, line 18: question: What do you think about settings indent to 2 spaces?

docs/building.md
```diff
@@ -1,130 +1,212 @@
...
+# Compilation options
 
-### Game client version
+## Game client version
 
 By default, the API assumes the latest version of the game client. The assumed version can be found in cmake's output, e.g.:
 ```bash
 $ cmake ../ | grep 'SC2 version'
-Target SC2 version: 5.0.5
+Target SC2 version: 5.0.15
```
alkurbatov, line 209: todo: Actually it is 5.0.14 (see set(SC2_VERSION "5.0.14" CACHE STRING "Version of the target StarCraft II client")). So should be changed there too.

src/sc2api/CMakeLists.txt
```diff
@@ -57,6 +57,27 @@ target_include_directories(sc2api SYSTEM PUBLIC "${civetweb_SOURCE_DIR}/include"
...
 target_link_libraries(sc2api PUBLIC sc2protocol civetweb-c-library)
 
+# Platform-specific system networking and threading libraries
+if(WIN32)
+    target_link_libraries(sc2api PUBLIC ws2_32 winmm)
+elseif(APPLE)
+    find_library(COREFOUNDATION_FRAMEWORK CoreFoundation)
+    target_link_libraries(sc2api PUBLIC ${COREFOUNDATION_FRAMEWORK})
+elseif(UNIX)
+    find_package(Threads REQUIRED)
```
alkurbatov, line 67: issue: This approach goes against the current idea adopted by the project: all find_package directives are called from the root CMakeLists.txt. 

Also, I doubt that we really need threads here.

src/sc2api/CMakeLists.txt
```diff
@@ -57,6 +57,27 @@ target_include_directories(sc2api SYSTEM PUBLIC "${civetweb_SOURCE_DIR}/include"
 
 target_link_libraries(sc2api PUBLIC sc2protocol civetweb-c-library)
 
+# Platform-specific system networking and threading libraries
+if(WIN32)
+    target_link_libraries(sc2api PUBLIC ws2_32 winmm)
+elseif(APPLE)
+    find_library(COREFOUNDATION_FRAMEWORK CoreFoundation)
```
alkurbatov, line 64: question: Could you please describe what is it and why it is linked with the API.

src/sc2api/CMakeLists.txt
```diff
@@ -57,6 +57,27 @@ target_include_directories(sc2api SYSTEM PUBLIC "${civetweb_SOURCE_DIR}/include"
...
+# Cross-platform compiler warnings and flags
 if (MSVC)
-    target_compile_options(sc2api PRIVATE /W4 /WX-)
-endif ()
+    target_compile_options(sc2api PRIVATE
+            /W4
+            /WX-
+    )
+else()
+    target_compile_options(sc2api PRIVATE
```
alkurbatov, line 78: suggestion: I think all such options should be moved to src/CMakeLists.txt (we have some in the root CMakeLists.txt too) for convenience and to keep everything in one place.

docs/building.md
```diff
@@ -134,23 +216,52 @@ always behind the Windows version. It is possible by specifying the game version
...
+```bash
+cmake --build --preset <preset-name> --parallel $(nproc --ignore=1)
+```
 
-If project compilation freezes, decrease nproc to 1 or more, e.g.:
+</details>
 
-``` bash
-$ cmake --build build --parallel $(nproc --ignore=1)
+Or just drop the number of cores to 1:
```
alkurbatov, line 252: issue: It means quite slow build. I doubt we should recommend it. 1 free CPU usually enough to solve this problem.

docs/building.md
```diff
@@ -134,23 +216,52 @@ always behind the Windows version. It is possible by specifying the game version
...
+# Troubleshooting
+
+## Build freezes
+
+If the final project build freezes (typically seen on underpowered machines), limit the parallel cores manually:
+
+Replace `<preset-name>` with the identifier found in your `CMakePresets.json` (e.g., `x64-debug`).
+
+
+<details name="build_freeze"><summary>Windows</summary>
```
alkurbatov, line 228: todo: Please don't mix in HTML. It could work in Github, but makes text harder to render and support, especially with different other non-github tools.

alkurbatov (changes requested):
Hi @spaceghost1589 !
Thank you for the PR!

README.md
spaceghost1589: The [benefits of Ninja](https://ninja-build.org/manual.html#_design_goals) include:
- Faster Build Times
- Incremental Building: Only changed files are rebuilt.
- Better Parallelism: Out-of-the-box performance without manual -j tuning.
- Better Dependency Handling: Strict and correct tracking.
- Modern Integration: Default support across popular environments.
- CMake Presets: Easily customizable by users for profile auto-population.

In relation to Linux and Makefiles:
- Lightweight: While it does require an additional tool on Linux, it's a single-binary package that's readily available on package managers (`sudo apt install ninja-build`).
- Backward Compatibility: Fully compatible with Makefiles. If a user prefers not to install Ninja, they can generate Makefiles instead: `cmake -B build -G "Unix Makefiles"`
- Cross-Platform Uniformity: Simplified CI/CD pipelines.

I'll be sure to add these commands and the benefits of Ninja to the build instructions.

cmake/toolchain/x86-64-w64-mingw32.cmake
spaceghost1589: Yes, I believe so. While a *native* Windows build typically reports `x86` or `AMD64` based on the host environment, this is a cross-compilation toolchain file used to build 64-bit Windows binaries from Linux.

Per the [CMake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html):

> When cross-compiling, a CMAKE_TOOLCHAIN_FILE should set the CMAKE_SYSTEM_PROCESSOR variable to match target architecture that it specifies (via CMAKE_&lt;LANG&gt;_COMPILER...).

Because both compiler specifications explicitly target `x86_64`, the CMAKE_SYSTEM_PROCESSOR should match that architecture:

set(CMAKE_C_COMPILER /usr/bin/**x86_64**-w64-mingw32-gcc)
set(CMAKE_CXX_COMPILER /usr/bin/**x86_64**-w64-mingw32-g++)

This [example](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-linux) shows the architecture matching as well.

docs/precompiled_libs.md
spaceghost1589: Done. Will be in update commit.

.clang-format
spaceghost1589: For JSON? Sounds good.

Honestly, I just tossed the JSON block in there so it would pass the linter, copy-pasting the C++ settings as a starting point. I figured you maintainers would want to tweak it to your preference, so I didn't spend any time on it. I'll go ahead and update the JSON IndentWidth to 2.

Let me know if you have any other formatting preferences and I'll split it off into its own PR; it might be a bit before I can redo the build instructions.

Side note on clang-tidy:
I noticed that clang-tidy was never fully implemented; it needs a line added to .pre-commit-config.yaml for the linter to pick up on it. Is there a plan to implement it?

There are quite a few warnings, so if implemented, the linter will likely fail until everything can be cleaned up (which I suspect is why it wasn't turned on in the first place). I can throw the .pre-commit-config.yaml change into a draft PR or open an Issue (or both) as a reminder and file tracker for pre-implementation clean-up if that's helpful.

docs/building.md
spaceghost1589: If reserving a single core for system functions is not enough consider:
- Ignoring additional cores, incrementing `1` to `#` ignored cores.
- Using a swap file for virtual RAM:

```bash
sudo fallocate -l 4G /swapfile
sudo chmod 600 /swapfile
sudo mkswap /swapfile
sudo swapon /swapfile
```

Delete swap file when finished:

```bash
sudo swapoff /swapfile
```

TODO Add Windows pagefile and macOS SSV instructions.

docs/building.md
spaceghost1589: Ahh man, I worked so hard on it too. :broken_heart:

Jk, I can take it out.

src/sc2api/CMakeLists.txt
spaceghost1589: These are system-level requirements pulled in networking layers like CivetWeb, which power the communication components in sc2server.cc and sc2connection.cc.

Ninja is fast because it's intentionally lean, meaning transitive dependencies don't always auto-propagate system libraries down to sc2api across every architecture and toolchain. Explicitly linking them ensures we have direct access. Is it strictly necessary in every single environment? Maybe, maybe not; you'd have to test it across multiple setups to find out, and it can vary by host architecture. But does it hurt? No, it prevents sudden build failures and guarantees cross-platform stability.

- ws2_32: Delivers the core networking API required for Windows applications to handle TCP/IP and socket-based communications.

- winmm (Windows Multimedia API): Legacy Windows services for multimedia management, audio control, and high-resolution timing.

- CoreFoundation: Provides foundational C-based data types and low-level system services for macOS.

src/sc2api/CMakeLists.txt
spaceghost1589: I created a reusable project_warnings interface target using generator expressions in root:

```
## Cross-platform compiler warnings and flags
# REUSABLE WARNING TARGET
add_library(project_warnings INTERFACE)
target_compile_options(project_warnings INTERFACE
    $<$<CXX_COMPILER_ID:MSVC>:/W4 /WX->
    $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-Wall -Wextra -Wno-unused-parameter>
)`
```

We can now simply link `project_warnings` to our maintainable code directories (src, tests, examples). This keeps things centralized and avoids warning spam from external dependencies (I tried implementing it globally; it was excessive).

src/sc2api/CMakeLists.txt
spaceghost1589: Moved all `find_package` and `find_library` up to root as specified.

Regarding Threads: it's the same [transitive dependency issue](https://github.com/cpp-sc2/cpp-sc2/pull/160#discussion_r3750195832) I described. Ninja's intentional leanness means transitive thread requirements don't always auto-propagate reliably across every Unix toolchain. Keeping Threads `target_link_libraries` ensures Unix builds don't fail on missing thread flags.

docs/building.md
spaceghost1589: I understand wanting to keep the core documentation focused on a clean, universal CLI setup (cmake + ninja) without promoting specific proprietary tools.

I'll go ahead and strip out the IDE-specific steps and implement CLI build instructions instead. The HTML as well.

docs/building.md
alkurbatov: Honestly, I would remove this recommendation completely. The swap part really system dependant and even on GNU Linux I know at least 3 implementations of it. So this could be simplified to "use --ignore " (as it was before) or "remove --parallel flag".

src/sc2api/CMakeLists.txt
alkurbatov: Looks great, thank you!

There is a nice example of this implementation here: https://github.com/cpp-best-practices/cmake_template/blob/main/cmake/CompilerWarnings.cmake which we can use to improve the code.

src/sc2api/CMakeLists.txt
alkurbatov: But currently we work without these libs so I wonder why do we need to link them now.

> But does it hurt? No, it prevents sudden build failures and guarantees cross-platform stability.

Well, it does, because it requires additional libraries that should be available somehow. If we get them implicitly from civetweb, it could be OK, because we don't use them directly.

I would suggest to go from particular compilation issues that require linking of additional components.

src/sc2api/CMakeLists.txt
alkurbatov: > Threads target_link_libraries ensures Unix builds don't fail on missing thread flags.

It doesn't fail on CI so it seems that we can work without it.

README.md
alkurbatov: The problem with this particular library implementation is that we build protobuf, SDL and civetweb most of the time (but never change their code). The library itself is quite small. So it seems that ninja provides almost no benefits for such a small project but instead introduces additional complexity.

I think ccache could help much more.

cmake/toolchain/x86-64-w64-mingw32.cmake
alkurbatov: Got it, thank you for detailed explanation.
Lets keep it.

alkurbatov:
Excuse me for such a long response. Life can be busy sometimes :(

.clang-format
alkurbatov: > I noticed that clang-tidy was never fully implemented; it needs a line added to .pre-commit-config.yaml for the linter to pick up on it. Is there a plan to implement it?

Yeah, this was TBD. Unfortunately requires a lot of changes to fully support strict config or a kind of 'diff after' check on CI.

I suggest to reformat the new code (e.g. presets). The rest will be fixed someday :)

spaceghost1589:
NP, I've been busy as well and haven't revisited this.

I'm closing this for now though. The scope got a bit too large, but the groundwork for modernizing the CMake build (Ninja + CMakePresets.json) and the Windows build guide is here. I may pull out smaller, focused pieces of this in the future.

PR #161, opened 2026-08-20
feat: remote Connect host, civetweb client-only init, host protoc when cross-compiling

ShadowLordAlpha:
Three pieces. Happy to split if that is easier to review.

## Coordinator::SetNetAddress

`Connect(int)` already reads `process_settings.net_address`, but there was no public setter, so the host was stuck at `127.0.0.1`. Call this before `Connect` when the websocket is not on localhost.

```cpp
coordinator.SetNetAddress(host);
coordinator.Connect(game_port);
coordinator.SetupPorts(2, start_port, false);
coordinator.JoinGame();
```

## Civetweb client-only init

cpp-sc2 only uses mg_connect_websocket_client. StartCivetweb() called mg_start2(), which starts an HTTP server and binds 8080 by default. That listen socket is then the process's first TCP port, so anything that maps the bot by pid (AI Arena get_ipv4_port_for_pid) never sees the real websocket source port.

This switches to mg_init_library(MG_FEATURES_WEBSOCKET). After that, Player Details, JoinGame, and RequestPing succeed against aiarena/arenaclient-bot v0.8.0.

## Host protoc when cross-compiling

Supersedes the CMake from #158 (Cross).

• Honor Protobuf_PROTOC_EXECUTABLE and CMAKE_CROSSCOMPILING so MinGW/Docker can generate protos with a host protoc instead of only WSL2_CROSS_COMPILE precompiled dumps.
• If cross-compiling and no executable was passed, search the host PATH (NO_CMAKE_FIND_ROOT_PATH) and fail with a clear error.
• Native builds still build and DEPENDS the in-tree protoc target.
• WSL2_CROSS_COMPILE is unchanged.
• Host protoc must match the protobuf GIT_TAG (still v33.0).

### DIFF
--- a/src/sc2api/sc2_connection.cc
+++ b/src/sc2api/sc2_connection.cc
@@ -15,23 +15,14 @@ bool StartCivetweb() {
         return true;
     }
 
-    const char* options[] = {
-        "request_timeout_ms", "5000", "websocket_timeout_ms", "1200000", "num_threads", "4", "tcp_nodelay", "1", 0};
-
-    struct mg_callbacks callbacks = {0};
-
-    struct mg_init_data mg_start_init_data = {0};
-    mg_start_init_data.callbacks = &callbacks;
-    mg_start_init_data.configuration_options = options;
-
-    struct mg_error_data mg_start_error_data = {0};
-    char ebuff[256] = {0};
-    mg_start_error_data.text = ebuff;
-    mg_start_error_data.text_buffer_size = sizeof(ebuff);
-
-    const auto* ctx = mg_start2(&mg_start_init_data, &mg_start_error_data);
-    if (!ctx) {
-        std::cerr << "Failed to start civetweb server: " << ebuff << std::endl;
+    // cpp-sc2 is a websocket client. mg_start2() starts an HTTP server that
+    // binds 8080 by default, so the process appears to listen on 8080 even
+    // though the SC2 connection uses an ephemeral source port. AI Arena's
+    // get_ipv4_port_for_pid then maps the bot to the listen port and never
+    // forwards the real websocket. mg_init_library only enables the client.
+    const unsigned got = mg_init_library(MG_FEATURES_WEBSOCKET);
+    if ((got & MG_FEATURES_WEBSOCKET) == 0) {
+        std::cerr << "Failed to init civetweb websocket client" << std::endl;
         return false;
     }
 
--- a/src/sc2api/sc2_coordinator.h
+++ b/src/sc2api/sc2_coordinator.h
@@ -70,8 +70,10 @@ public:
     //! \param port_start First port number.
     void SetPortStart(int port_start);
 
-    //! Sets the address used by Connect(int). Required for AI Arena `--LadderServer`.
-    //! \param net_address Host to attach to (for example a compose service IP).
+    //! Sets the host used by Connect(int). Default is 127.0.0.1.
+    //! Needed when attaching to a remote SC2 websocket (AI Arena `--LadderServer`).
+    //! \param net_address Hostname or IP to attach to.
+    //! \sa Connect
     void SetNetAddress(const std::string& net_address);
 
     //! Indicates whether feature layers should be provided in the observation.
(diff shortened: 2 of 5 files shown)


src/sc2api/sc2_coordinator.h
```diff
@@ -70,6 +70,12 @@ class Coordinator {
     //! \param port_start First port number.
     void SetPortStart(int port_start);
 
+    //! Sets the host used by Connect(int). Default is 127.0.0.1.
+    //! Needed when attaching to a remote SC2 websocket (AI Arena `--LadderServer`).
+    //! \param net_address Hostname or IP to attach to.
+    //! \sa Connect
+    void SetNetAddress(const std::string& net_address);
```
alkurbatov, line 77: suggestion: Could be std:string_view or moved std::string.

src/sc2protocol/CMakeLists.txt
```diff
@@ -44,6 +44,26 @@ if (WSL2_CROSS_COMPILE)
     endforeach ()
 # Compile protos with protoc
 else ()
+    if (Protobuf_PROTOC_EXECUTABLE)
```
alkurbatov: question: Could you please describe the CMAKE_CROSSCOMPILING case. We already have WSL, so it seems that there is no need in another cross-mode.

src/sc2protocol/CMakeLists.txt
```diff
@@ -44,6 +44,26 @@ if (WSL2_CROSS_COMPILE)
...
+        set(protoc_executable ${Protobuf_PROTOC_EXECUTABLE})
+        set(protoc_depends ${Protobuf_PROTOC_EXECUTABLE})
```
alkurbatov: suggestion: I think this code should live in thirdparty/cmake/protobuf.cmake This module already provides some variables regarding protoc.

src/sc2api/sc2_connection.cc
```diff
@@ -15,23 +15,14 @@ bool StartCivetweb() {
...
-    const auto* ctx = mg_start2(&mg_start_init_data, &mg_start_error_data);
-    if (!ctx) {
-        std::cerr << "Failed to start civetweb server: " << ebuff << std::endl;
+    // cpp-sc2 is a websocket client. mg_start2() starts an HTTP server that
+    // binds 8080 by default, so the process appears to listen on 8080 even
+    // though the SC2 connection uses an ephemeral source port. AI Arena's
+    // get_ipv4_port_for_pid then maps the bot to the listen port and never
+    // forwards the real websocket. mg_init_library only enables the client.
+    const unsigned got = mg_init_library(MG_FEATURES_WEBSOCKET);
+    if ((got & MG_FEATURES_WEBSOCKET) == 0) {
```
alkurbatov, line 24: question: Looks like it fixes #133?

alkurbatov (changes requested):
Thank you for the patch!

Overall implementation is good.
I have several question regarding the implementation and minor suggestion.

src/sc2api/sc2_coordinator.h
ShadowLordAlpha: Every other string in coordinator is `const std::string&` I can change it really needed but it would then be the odd one out

src/sc2api/sc2_connection.cc
ShadowLordAlpha: It actually should fix that as far as I can tell, I didn't bump the civetweb version though so didn't check that one specifically, but it did look like the same issue to me

What happened was an HTTP server was getting started that was never used, and that also was causing the process to bind port 8080 which AIArena got confused by and was trying to connect to. If you don't start the HTTP server the issue goes away

src/sc2protocol/CMakeLists.txt
ShadowLordAlpha: Yes I can move it

src/sc2protocol/CMakeLists.txt
ShadowLordAlpha: The WSL path only does precompiled-proto, the Protobuf_PROTOC_EXECUTABLE lets me use a linux version of protoc instead which is helpful when im not using WSL to compile my code (Docker mainly, so Linux making a windows exe instead of just using WSL which i find is really annoying to make portable).

CMAKE_CROSSCOMPILING is basically just a cmake flag that is enabled when building for a different system. So while its true when doing WSL, its also true for building for Windows when on Linux

src/sc2protocol/CMakeLists.txt
ShadowLordAlpha: This has now been moved

src/sc2protocol/CMakeLists.txt
alkurbatov: Ah I see, makes sense.

src/sc2api/sc2_coordinator.h
alkurbatov: Fair enough, ok, lets keep it as is.

alkurbatov (approved):
Thank you!
Looks good to me.

PR #162, opened 2026-09-07
feat: copy RequestReplayInfo player names into ReplayPlayerInfo

### LINKED ISSUE #25: ReplayInfo is not filled in by ReplayControlImp::GatherReplayInfo
Running the example program replay.cc fails. 

The program finds the replay folder and verifies the game file. The data for the game file is present (based on the fact that I can watch the replay from the sc2 game itself). 
But, the request to the protobuf for replay information is empty.  The return value and all existing tests in the code suggest that the request is successful (e.g. response->error_size() == 0 && has_replay_info() == true). 
But, ResponseReplayInfo structure, which is entirely optional fields, has no data. That is, calls to has_fieldname() all return false.
Despite there being no overt error caught by the program. All is not well, because if I examine the response from protobuf (in ControlImp::WaitForResponse(), look at response.get().DebugString()), it says
```
 error: ParsingError
  error_details: "Could not open initData for the replay: C:/Replays/03032019 - GAME 1 - soO vs Stats - Z vs P - Automaton LE.SC2Replay"
```
What's also weird, is that normally the example will skip replays if the duration is too short. In this case, the error makes it appear to be 0 frames, but if I I force it to go on, the replay actually works, but then, the OnStep() function of the ReplayObserver are not called. Is it supposed to be?

Other example programs, like annoying helper, appear to work.

ShadowLordAlpha:
Fixes #25.

## Why
Issue #25 reported GatherReplayInfo succeeding while ReplayInfo looked empty
(duration 0, example skipped the file, ReplayObserver::OnStep never ran).
Most optional-field copying is already on master. The remaining gap is player
names: ResponseGameInfo during playback often leaves PlayerInfo.player_name
empty, and ReplayPlayerInfo did not copy RequestReplayInfo's names either.

## What
- Copy player_info().player_name() from ResponseReplayInfo into ReplayPlayerInfo.name.
- test_replay_observation captures a short same-version 1v1 on Bel'Shir Vestige LE
  (HelixAlpha Terran vs HelixBravo Zerg), saves under maps/ReplayFixtures/, then
  loads it through SetReplayPath + ReplayObserver.
- Success requires IgnoreReplay's ReplayInfo and that OnStep ran. Camera/chat are
  recorded in the fixture but are not pass/fail on this branch.

AssertReplayInfoComplete fails the test unless all of the following hold:

- num_players == 2
- map_name, replay_path, version, data_version non-empty
- duration > 0 and duration_gameloops != 0
- data_build != 0 and base_build != 0
- players include named Terran HelixAlpha and named Zerg HelixBravo
  (name plus race / race_selected)

scan_directory joins directory + filename with a separator so
RequestReplayInfo.replay_path matches the file SaveReplay wrote.

## Tested
Live SC2 (Windows, Base97563). ReplayInfo had both names, non-zero duration/builds, map/version strings, and
the observer stepped.

### DIFF
--- a/src/sc2api/sc2_replay_observer.cc
+++ b/src/sc2api/sc2_replay_observer.cc
@@ -112,6 +112,9 @@ bool ReplayControlImp::GatherReplayInfo(const std::string& path, bool download_d
         if (player_info_extra_proto.has_player_info()) {
             const SC2APIProtocol::PlayerInfo& player_info_proto = player_info_extra_proto.player_info();
             player_info.player_id = player_info_proto.player_id();
+            if (player_info_proto.has_player_name()) {
+                player_info.name = player_info_proto.player_name();
+            }
             if (player_info_proto.has_race_actual()) {
                 player_info.race = ConvertRaceFromProto(player_info_proto.race_actual());
             }
--- a/src/sc2utils/sc2_scan_directory.cc
+++ b/src/sc2utils/sc2_scan_directory.cc
@@ -13,6 +13,23 @@
 
 namespace sc2 {
 
+namespace {
+
+std::string JoinDirFile(const char* directory_path, const char* name) {
+    std::string path(directory_path);
+    if (!path.empty() && path.back() != '/' && path.back() != '\\') {
+#ifdef _WIN32
+        path += '\\';
+#else
+        path += '/';
+#endif
+    }
+    path += name;
+    return path;
+}
+
+}  // namespace
+
 int scan_directory(const char* directory_path, std::vector<std::string>& files, bool full_path, bool list_directories) {
     if (!directory_path || !*directory_path) {
         return 0;
@@ -36,8 +53,9 @@ int scan_directory(const char* directory_path, std::vector<std::string>& files, 
                 if (!full_path) {
                     files.push_back(ent->d_name);
                 } else {
-                    files.push_back(std::string(directory_path) + std::string(ent->d_name));
+                    files.push_back(JoinDirFile(directory_path, ent->d_name));
                 }
+                break;
             }
             case DT_DIR: {
                 if (!list_directories || !*ent->d_name) {
@@ -51,9 +69,10 @@ int scan_directory(const char* directory_path, std::vector<std::string>& files, 
                 if (!full_path) {
                     files.push_back(ent->d_name);
                 } else {
-                    files.push_back(std::string(directory_path) + std::string(ent->d_name));
+                    files.push_back(JoinDirFile(directory_path, ent->d_name));
                 }
-            } break;
+                break;
+            }
         }
     }
 
--- a/tests/test_replay_observation.cc
+++ b/tests/test_replay_observation.cc
@@ -139,7 +139,6 @@ public:
         } else if (saw_chat_text_) {
             chat_cleared_ = true;
         }
-
     }
 
     bool stepped() const {
--- a/tests/test_replay_observation.cc
+++ b/tests/test_replay_observation.cc
@@ -1,0 +1,234 @@
+#include "test_replay_observation.h"
+
+#include <filesystem>
+#include <iostream>
+#include <string>
+
+#include "sc2api/sc2_api.h"
+#include "sc2lib/sc2_lib.h"
+#include "sc2utils/sc2_manage_process.h"
+
+namespace sc2 {
+
+namespace {
+
+const char* kPlayerAlpha = "HelixAlpha";
+const char* kPlayerBravo = "HelixBravo";
+const char* kChatLine = "cpp-sc2 replay chat probe";
+const Point2D kCameraTarget(24.0f, 28.0f);
+
+void Report(std::string* errors, const std::string& message) {
+    std::cerr << "TestReplayObservation: " << message << std::endl;
+    if (errors->empty()) {
+        *errors = message;
+    } else {
+        *errors += "; " + message;
+    }
+}
+
+bool HasPlayerNamed(const ReplayInfo& info, const std::string& name, Race expected_race) {
+    for (int i = 0; i < info.num_players; ++i) {
+        const ReplayPlayerInfo& player = info.players[i];
+        if (player.name == name) {
+            return player.race == expected_race || player.race_selected == expected_race;
+        }
+    }
+    return false;
+}
+
+bool AssertReplayInfoComplete(const ReplayInfo& info, const std::string& replay_path, std::string* errors) {
+    bool ok = true;
+    if (info.num_players != 2) {
+        Report(errors, "expected 2 players, got " + std::to_string(info.num_players));
+        ok = false;
+    }
+    if (info.map_name.empty()) {
+        Report(errors, "map_name is empty");
+        ok = false;
+    }
+    if (info.replay_path.empty()) {
+        Report(errors, "replay_path is empty");
+        ok = false;
+    }
+    if (info.version.empty()) {
+        Report(errors, "version is empty");
+        ok = false;
+    }
+    if (info.data_version.empty()) {
+        Report(errors, "data_version is empty");
+        ok = false;
+    }
+    if (info.duration <= 0.0f) {
+        Report(errors, "duration is not positive");
+        ok = false;
+    }
+    if (info.duration_gameloops == 0) {
+        Report(errors, "duration_gameloops is 0");
+        ok = false;
+    }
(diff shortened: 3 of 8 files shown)


src/sc2utils/sc2_scan_directory.cc
```diff
@@ -13,6 +13,23 @@
 
 namespace sc2 {
 
+namespace {
+
+std::string JoinDirFile(const char* directory_path, const char* name) {
+    std::string path(directory_path);
+    if (!path.empty() && path.back() != '/' && path.back() != '\\') {
```
alkurbatov: todo: Please use std::filesystem to solve separators issue.

E.g.

```cpp
std::filesystem::path p(directory_path);
p /= name;
return p.string();
```

src/sc2utils/sc2_scan_directory.cc
```diff
@@ -13,6 +13,23 @@
 
 namespace sc2 {
 
+namespace {
+
+std::string JoinDirFile(const char* directory_path, const char* name) {
```
alkurbatov: nitpick: Please avoid raw pointers if possible. We can use std::string_view or references to std::string here.

src/sc2api/sc2_replay_observer.cc
```diff
@@ -112,6 +112,9 @@ bool ReplayControlImp::GatherReplayInfo(const std::string& path, bool download_d
         if (player_info_extra_proto.has_player_info()) {
             const SC2APIProtocol::PlayerInfo& player_info_proto = player_info_extra_proto.player_info();
             player_info.player_id = player_info_proto.player_id();
+            if (player_info_proto.has_player_name()) {
+                player_info.name = player_info_proto.player_name();
```
alkurbatov: question: Probably just

```c++
player_info.name = player_info_proto.player_name();
```
?

tests/test_replay_observation.cc
```diff
@@ -0,0 +1,238 @@
...
+
+#include "sc2api/sc2_api.h"
+#include "sc2lib/sc2_lib.h"
+#include "sc2utils/sc2_manage_process.h"
+
+namespace sc2 {
+
+namespace {
+
+const char* kPlayerAlpha = "HelixAlpha";
```
alkurbatov: nitpick:

Please use constexpr for such constants. E.g.:
```cpp
constexpr std::string_view kPlayerAlpha = "HelixAlpha";
```

Here and below.

tests/test_replay_observation.cc
```diff
@@ -0,0 +1,238 @@
...
+    std::cerr << "TestReplayObservation: " << message << std::endl;
+    if (errors->empty()) {
+        *errors = message;
+    } else {
+        *errors += "; " + message;
+    }
+}
+
+bool HasPlayerNamed(const ReplayInfo& info, const std::string& name, Race expected_race) {
+    for (int i = 0; i < info.num_players; ++i) {
```
alkurbatov: suggestion: Seems that std::find_if perfectly fits here, e.g.:

```c++
bool HasPlayerNamed(const ReplayInfo& info, const std::string& name, Race expected_race) {
    const auto begin = info.players;
    const auto end = info.players + info.num_players;
    const auto it = std::find_if(begin, end, [&name](const ReplayPlayerInfo& player) {
        return player.name == name;
    });
    return it != end && (it->race == expected_race || it->race_selected == expected_race);
}
```

tests/test_replay_observation.cc
```diff
@@ -0,0 +1,238 @@
...
+    }
+    if (!HasPlayerNamed(info, kPlayerAlpha, Race::Terran)) {
+        Report(errors, std::string("missing named Terran player ") + kPlayerAlpha);
+        ok = false;
+    }
+    if (!HasPlayerNamed(info, kPlayerBravo, Race::Zerg)) {
+        Report(errors, std::string("missing named Zerg player ") + kPlayerBravo);
+        ok = false;
+    }
+    (void)replay_path;
```
alkurbatov: nitpick: I suggest to drop this arg as it is not used in the code. If you think it should stay here, it would be better to use either static_cast (as we are in C++), or leave the arg without a name.

tests/test_replay_observation.cc
```diff
@@ -0,0 +1,238 @@
...
+namespace sc2 {
+
+namespace {
+
+const char* kPlayerAlpha = "HelixAlpha";
+const char* kPlayerBravo = "HelixBravo";
+const char* kChatLine = "cpp-sc2 replay chat probe";
+const Point2D kCameraTarget(24.0f, 28.0f);
+
+void Report(std::string* errors, const std::string& message) {
```
alkurbatov: suggestion: A bit more modern solution:

```c++
void Report(std::ostringstream& errors, std::string_view message) {
    std::cerr << "TestReplayObservation: " << message << std::endl;

    if (errors.tellp() > 0) {
        errors << "; ";
    }

    errors << message;
}
```

tests/test_replay_observation.cc
```diff
@@ -0,0 +1,238 @@
...
+            std::cerr << "TestReplayObservation: SaveReplay failed at " << replay_path << " size=" << size
+                      << " loop=" << bot.Observation()->GetGameLoop() << std::endl;
+            return false;
+        }
+    }
+
+    std::string errors;
+    Coordinator replay_coordinator;
+    if (!replay_coordinator.LoadSettings(argc, argv)) {
+        std::cerr << "TestReplayObservation: replay LoadSettings failed" << std::endl;
```
alkurbatov: nitpick: Slightly faster and shorter:
<< std::endl -> << '\n' or even std::cerr << "TestReplayObservation: replay LoadSettings failed\n";

Here and in other places.

tests/test_replay_observation.cc
```diff
@@ -0,0 +1,238 @@
...
+    if (!observer.info_ok()) {
+        std::cerr << "TestReplayObservation: ReplayInfo incomplete: " << errors << std::endl;
+    }
+    if (!observer.stepped()) {
+        std::cerr << "TestReplayObservation: ReplayObserver never stepped" << std::endl;
+        success = false;
+    }
+    // Camera and chat assertions land in later topic PRs. This test only
+    // requires a complete ReplayInfo including both player names.
+    (void)observer.saw_chat_text();
```
alkurbatov: question: Looks like a dead code.

alkurbatov (changes requested):
Hey @ShadowLordAlpha, thank you for the patch!

I left several suggestions regarding the implementation.

alkurbatov:
Lets finish this patch first as it has many common commits with other patches.

ShadowLordAlpha:
> Lets finish this patch first as it has many common commits with other patches.

Ya this one turned out to be kinda a base for the others I posted (not including the draft ones I have to go through and check again in a day or so) I tried to prevent that in the others honestly

src/sc2api/sc2_replay_observer.cc
ShadowLordAlpha: The proto has that method so I was using it, I'll drop from the test as it should auto fill in with just ""

src/sc2utils/sc2_scan_directory.cc
ShadowLordAlpha: Adjusted, now uses filesystem

src/sc2utils/sc2_scan_directory.cc
ShadowLordAlpha: I swapped it to a filesystem and string_view

tests/test_replay_observation.cc
ShadowLordAlpha: Done

tests/test_replay_observation.cc
ShadowLordAlpha: Done

tests/test_replay_observation.cc
ShadowLordAlpha: Done

tests/test_replay_observation.cc
ShadowLordAlpha: Done

tests/test_replay_observation.cc
ShadowLordAlpha: done

tests/test_replay_observation.cc
ShadowLordAlpha: removed

alkurbatov (approved):
Thank you!
Nice work.

PR #163, opened 2026-09-07
feat: expose raw camera moves in observation actions

### LINKED ISSUE #102: Add raw camera moves to observation
https://github.com/cpp-sc2/cpp-sc2/pull/99

ShadowLordAlpha:
Fixes #102.

## Why
ConvertRawActions skipped raw camera pans, so replay and DebugMoveCamera
moves never showed up in observations. Stuffing them into ActionRaw left
ability/unit fields unused.

## What
- RawCameraMove (Point3D world-space center) on GetRawCameraMoves().
- OnCameraMove for each pan this step (IssueEvents, before OnStep).
- ActionRaw stays unit commands only.
- echo_actions prints camera overlay when no unit command is echoed.
- test_replay_observation pans to playable center, asserts live and
  recorded raw pans, and checks spectator ObserverAction::CameraMove does
  not appear as a raw move.

Not in this PR: spectator camera pose. The protocol has no observer camera
in Observation, and ResponseObserverAction is empty.

## Tested
Live SC2 (Windows, Base97563), Bel'Shir Vestige LE. TestReplayObservation
passed: DebugMoveCamera seen live, recorded pan in replay, spectator pan
does not show up in GetRawCameraMoves.

### DIFF
--- a/src/sc2api/sc2_action.h
+++ b/src/sc2api/sc2_action.h
@@ -19,7 +19,9 @@ struct ActionRaw {
         //! The target is a unit tag, could also be a snapshot in the fog-of-war.
         TargetUnitTag,
         //! The target is a point.
-        TargetPosition
+        TargetPosition,
+        //! Raw camera move. ability_id is unused; target_point/target_z hold world space.
+        TargetCamera
     };
 
     //! The ID of the ability to invoke.
@@ -30,8 +32,10 @@ struct ActionRaw {
     TargetType target_type = TargetNone;
     //! The target of this action. Valid only when target_type == TargetUnitTag.
     Tag target_tag = NullTag;
-    //! The target point for this action. Valid only when target_type == TargetPosition.
+    //! The target point for this action. Valid when target_type == TargetPosition or TargetCamera.
     Point2D target_point;
+    //! World-space Z for TargetCamera. Unused for other target types.
+    float target_z = 0.0f;
 
     //! Comparison overload.
 
@@ -49,6 +53,9 @@ struct ActionRaw {
             return false;
         }
         if (target_point.y != a.target_point.y) {
+            return false;
+        }
+        if (target_z != a.target_z) {
             return false;
         }
         return true;
--- a/src/sc2api/sc2_proto_to_pods.cc
+++ b/src/sc2api/sc2_proto_to_pods.cc
@@ -370,6 +370,18 @@ void ConvertRawActions(const ResponseObservationPtr& response_observation_ptr, R
             continue;
         }
         const SC2APIProtocol::ActionRaw& action_raw = proto_action.action_raw();
+        if (action_raw.has_camera_move()) {
+            const SC2APIProtocol::ActionRawCameraMove& camera_move = action_raw.camera_move();
+            ActionRaw action;
+            action.target_type = ActionRaw::TargetCamera;
+            if (camera_move.has_center_world_space()) {
+                action.target_point.x = camera_move.center_world_space().x();
+                action.target_point.y = camera_move.center_world_space().y();
+                action.target_z = camera_move.center_world_space().z();
+            }
+            actions.push_back(action);
+            continue;
+        }
         if (!action_raw.has_unit_command()) {
             continue;
         }
--- a/src/sc2utils/sc2_scan_directory.cc
+++ b/src/sc2utils/sc2_scan_directory.cc
@@ -13,6 +13,23 @@
 
 namespace sc2 {
 
+namespace {
+
+std::string JoinDirFile(const char* directory_path, const char* name) {
+    std::string path(directory_path);
+    if (!path.empty() && path.back() != '/' && path.back() != '\\') {
+#ifdef _WIN32
+        path += '\\';
+#else
+        path += '/';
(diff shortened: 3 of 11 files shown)


src/sc2utils/sc2_scan_directory.cc
```diff
@@ -13,6 +13,23 @@
 
 namespace sc2 {
 
+namespace {
+
+std::string JoinDirFile(const char* directory_path, const char* name) {
```
alkurbatov, line 18: This requires rebase to the latest master.

alkurbatov (changes requested):

src/sc2api/sc2_action.h
```diff
@@ -30,8 +32,10 @@ struct ActionRaw {
     TargetType target_type = TargetNone;
     //! The target of this action. Valid only when target_type == TargetUnitTag.
     Tag target_tag = NullTag;
-    //! The target point for this action. Valid only when target_type == TargetPosition.
+    //! The target point for this action. Valid when target_type == TargetPosition or TargetCamera.
     Point2D target_point;
+    //! World-space Z for TargetCamera. Unused for other target types.
+    float target_z = 0.0f;
```
alkurbatov, line 38: todo: Seems that this one should be united with Point2D into Point3D.

src/sc2api/sc2_action.h
alkurbatov, line 42: suggestion: I believe we don't need comparison operator here as all of these are simple types.

src/sc2api/sc2_proto_to_pods.cc
```diff
@@ -370,6 +370,18 @@ void ConvertRawActions(const ResponseObservationPtr& response_observation_ptr, R
...
+        if (action_raw.has_camera_move()) {
+            const SC2APIProtocol::ActionRawCameraMove& camera_move = action_raw.camera_move();
+            ActionRaw action;
+            action.target_type = ActionRaw::TargetCamera;
+            if (camera_move.has_center_world_space()) {
+                action.target_point.x = camera_move.center_world_space().x();
+                action.target_point.y = camera_move.center_world_space().y();
+                action.target_z = camera_move.center_world_space().z();
+            }
+            actions.push_back(action);
```
alkurbatov, line 382: nitpick (perf):
Should be actions.push_back(std::move(action));

src/sc2api/sc2_action.h
alkurbatov, line 13: note: I am quite not sure that we should expose camera inside the Action structure. Yes, on the proto-level it is some action after all, but camera is not about unit or ability. As a result some fields of this struct cannot be filled properly which means bad design.

A separate vector of dedicated objects could solve this issue.

src/sc2api/sc2_action.h
```diff
@@ -19,7 +19,9 @@ struct ActionRaw {
         //! The target is a unit tag, could also be a snapshot in the fog-of-war.
         TargetUnitTag,
         //! The target is a point.
-        TargetPosition
+        TargetPosition,
+        //! Raw camera move. ability_id is unused; target_point/target_z hold world space.
+        TargetCamera
```
alkurbatov, line 24: todo: Since this code don't have tests I suggest to extend examples/echo_actions.cc to echo camera moves.

ShadowLordAlpha:
I'll go through the requested changes when i have a bit of free time. I assume you want them as something like RawCameraMove by themselves instead of part of RawAction right?

alkurbatov:
> I'll go through the requested changes when i have a bit of free time. I assume you want them as something like RawCameraMove by themselves instead of part of RawAction right?

Yeah, kind of. Allows to work with camera directly (probably needs additional callback) and avoid partly initialized object. See https://github.com/cpp-sc2/cpp-sc2/pull/99 it has similar approach.

src/sc2api/sc2_proto_to_pods.cc
ShadowLordAlpha: ConvertRawCameraMoves uses push_back(std::move(move)).

src/sc2api/sc2_action.h
ShadowLordAlpha: Dedicated RawCameraMove list plus OnCameraMove. Did not take PR #99’s shared_ptr / dynamic_cast.

src/sc2api/sc2_action.h
ShadowLordAlpha: OnCameraMove fills the overlay. If a unit command is present in the same step, this example still shows that command; documented on the handler.

ShadowLordAlpha:
> > I'll go through the requested changes when i have a bit of free time. I assume you want them as something like RawCameraMove by themselves instead of part of RawAction right?
> 
> Yeah, kind of. Allows to work with camera directly (probably needs additional callback) and avoid partly initialized object. See #99 it has similar approach.

Done. RawCameraMove is its own type on GetRawCameraMoves(), and
OnCameraMove fires from IssueEvents.

That is the player/debug pan (DebugMoveCamera / recorded raw pans),
not ObserverAction::CameraMove. Spectator pose is not in Observation,
and ResponseObserverAction is empty, so there is nothing to hook there.

PR #164, opened 2026-09-07
fix: skip tick on missing observation without crashing

### LINKED ISSUE #62: Prevent crash when observation is missing

ShadowLordAlpha:
Fixes #62.

## Why
GetObservation could receive a leftover `ResponseStep` (proto case 12) when waiting for Observation. The library cleared `response_pending_` after any receive, so the next tick sent another Step. WaitStep then treated the old Step as this tick's, and Observation consumed the new Step. Downstream OnStep crashed on a missing observation.

## What
- Stamp `Request.id` on send; wait for matching type and id (`MatchPendingResponse`).
- Drop stale responses (`id` older than pending). Same id with the wrong type is still `ResponseMismatch`.
- Do not send Observation while a response is still owed.
- On a failed GetObservation, keep the last good observation and skip OnStep. No ready flag, no extra library logs, no `DebugString()` dumps.
- `TestProtoMatch` covers the matcher without launching SC2.

## Tested
sc2api and all_tests compile (Ninja+MSVC). `TestProtoMatch` succeeded. Remaining all_tests entries still require a live client.


src/sc2api/sc2_client.cc
```diff
@@ -1991,45 +2014,90 @@ bool ControlImp::HasResponsePending() const {
...
 }
 
+bool ControlImp::IsObservationReady() const {
+    return observation_ready_;
+}
+
 bool ControlImp::GetObservation() {
     if (app_state_ != AppState::normal) {
+        observation_ready_ = false;
+        std::cerr << "[cpp-sc2 warning] GetObservation skipped: app_state=" << static_cast<int>(app_state_)
```
alkurbatov, line 2024: todo: Thats a lot of logs for a library. Ideally we should avoid it and either provide some data to get current state (e.g. list of errors), or get a logger instance provided by user to log into it. The problem with std::cerr is that it very slow and cannot be controlled by the user.

src/sc2api/sc2_client.cc
```diff
@@ -1991,45 +2014,90 @@ bool ControlImp::HasResponsePending() const {
...
             }
         } else {
-            std::cerr << "No error strings in result." << std::endl;
+            std::cerr << "  No error strings in result." << std::endl;
+        }
+        if (response->has_status()) {
+            std::cerr << "  proto_status=" << response->status() << std::endl;
         }
-        std::cerr << std::endl;
+        std::string dump = response->DebugString();
```
alkurbatov, line 2068: todo: Debug string should be remove as it is relatively slow.

src/sc2api/sc2_client.cc
```diff
@@ -1991,45 +2014,90 @@ bool ControlImp::HasResponsePending() const {
...
+                dump += "\n... [truncated]";
+            }
+            std::cerr << "  proto:\n" << dump << std::endl;
+        }
         return false;
     }
 
     observation_ = observation;
     response_ = response_observation;
+    observation_ready_ = true;
```
alkurbatov, line 2100: What do you think about switching to different approach:

```cpp
static Observation* last_obs = nullptr;
if (observation.HasErrors()) {
   // do something to set previous observation as current.
   return true; // probably shouldn't return anything in this function.
}

last_obs = response_observation;
observation_imp_->UpdateObservation();
```

This allows to store data close to it's source and avoid flags management which state should be synced with observation retrieval results.

ShadowLordAlpha:
so, im digging into this quite a bit more atm, and i think it might actually be something different. Speciically i think we may be getting out of order somehow.

What it looks like is actually happening
WaitForResponseInternal records ResponseMismatch on a wrong type, then still returns that proto and clears pending.
2. WaitForResponse treats any error_size() == 0 body as success, including a non-Step.
3. WaitStep sees !has_step(), returns false, and does not call GetObservation. The real Step is still queued.
4. Next tick Step() sends another Step (pending is empty, so send is allowed). WaitStep pops the old Step (has_step() succeeds) and GetObservation sends Observation. The new Step is next on the wire → Observation wait gets type 12.

PR #169, opened 2026-09-13
feat: calculate expansion locations from placement grid (#72)

### LINKED ISSUE #72: Improve expansions detection code
The current implementation of expansion detection algorithm is not precise and pretty slow.
Apply the same approach as in https://github.com/BurnySc2/python-sc2/blob/develop/sc2/bot_ai.py#L301

ShadowLordAlpha:
Fixes #72.

## Why
`CalculateExpansionLocations` walked a circle of `Query()->Placement` calls (~100ms) and was not very precise. Placement and height are already on `GameInfo`.

## What
- Find expansions from the placement/height grids with a 5x5 town-hall footprint. No `Query`.
- `Cluster()` is generic closest-pair grouping (position/tag tie-break). Expansion policy (height split, groups larger than 12, opposite-geyser layouts) stays in `CalculateExpansionLocations`.
- Resources come from `UnitTypeData::has_minerals` / `has_vespene`.
- Breaking: `CalculateExpansionLocations` no longer takes `QueryInterface*`; `radiuses_` / `circle_step_size_` are removed; `Cluster()` returns `std::vector<Units>`.

## Tested
- Live `TestExpansionLocations` on Bel'Shir Vestige LE: 10 expansions (hand-checked), official start locations match, `Query()->Placement` neighborhood check on unoccupied bases.
- Timing on that map: ~6.2 ms cold, ~0.53 ms after (5 runs).
- `sc2lib` / `all_tests` compile on Windows (VS 2026).

### DIFF
--- a/src/sc2lib/sc2_search.cc
+++ b/src/sc2lib/sc2_search.cc
@@ -1,10 +1,23 @@
 #include "sc2_search.h"
 
+#include <algorithm>
 #include <cmath>
+#include <limits>
+#include <optional>
+#include <utility>
+#include <vector>
+
+#include "sc2api/sc2_map_info.h"
+#include "sc2api/sc2_typeenums.h"
+#include "sc2api/sc2_unit_filters.h"
 
 namespace {
 const float PI = 3.1415927F;
-}
+// python-sc2 compares raw 0-255 height bytes with delta <= 10 (~1.25 world units).
+const float kHeightMergeDelta = 10.0F / 8.0F;
+const int kOffsetRange = 7;
+const size_t kMaxResourcesPerExpansion = 12;
+}  // namespace
 
 namespace sc2::search {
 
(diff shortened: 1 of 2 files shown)


spaceghost1589:
Impressive. 

Although you might consider: 

- `GatherExpansionResources` -  Add a minimum mineral amount to ensure mineral walls don't get included in the data.
- `ExpansionOffsets` - Opt for `DistanceSquared` instead of `hypot`.

ShadowLordAlpha:
> Impressive.
> 
> Although you might consider:
> 
>     * `GatherExpansionResources` -  Add a minimum mineral amount to ensure mineral walls don't get included in the data.
> 
>     * `ExpansionOffsets` - Opt for `DistanceSquared` instead of `hypot`.

We cannot see the amount of minerals in a base we have no vision of, though we already filter those that have more than 12 fields, I could also add a check for a full field UnitType though as the blockers only use the 450 type.

For the Offset there really isn't a benefit to do so unless you are calling that method every step or more. Though knowing some authors that would actually be the case...

PR #170, opened 2026-09-13
Optimize sc2protocol: Enable packed encoding and arena allocation

spaceghost1589:
### Description

This PR introduces a CMake patch to the sc2protocol dependency, modifying the sc2protocol  .proto files to enable two Protobuf optimizations: Packed Encoding and Arena Allocation.

These additions improve memory management and processing speed.

### Benefits

1. [Packed Encoding for Repeated Fields](https://protobuf.dev/programming-guides/proto2/) (`[packed=true]`)

    **What it does:** Forces arrays of primitive types (like integers and floats) to be stored in a contiguous block under a single tag, rather than writing a tag and wire type for every single element.

   **Benefit:** ~~It reduces the overall payload size of serialized data. This means a smaller memory footprint and faster parsing/network I/O, as the parser spends less time reading redundant tag bytes for large arrays.~~ *(See discussion below... Only applies to fields in messages we construct and send ourselves; three fields actually benefit.)*

2. [Arena Allocation](https://protobuf.dev/reference/cpp/arenas/) (`cc_enable_arenas = true;`)

    **What it does:** Allows C++ Protobuf objects to be allocated on an "Arena" (a pre-allocated, contiguous chunk of memory) rather than using standard heap allocation.

    **Benefit:** ~~It virtually eliminates the CPU overhead caused by individual new and delete calls when constructing and destroying complex, deeply-nested protocol buffers. De-allocating an entire message tree becomes instantaneous (the arena is just dropped), which completely prevents heap fragmentation and massively speeds up execution during heavy data loads.~~ *(See discussion below... This flag alone does nothing; requires messages to be explicitly constructed via google::protobuf::Arena.)*

3. Safe for all OSs and Toolchains

This patch is purely an instruction to the Protobuf compiler (protoc) and is fully cross-platform:

No OS or Compiler Dependencies: This doesn't introduce platform-specific C++ macros, system headers, or compiler flags. The Protobuf compiler natively handles these .proto options and generates standard, compliant C++ code equally on Windows, Linux, and macOS.

Wire Compatibility: Packed encoding is natively understood by all modern Protobuf parsers. It is strictly a wire-format optimization and will not break backward compatibility with existing data streams or replays.


ShadowLordAlpha:
Would this change even do anything? As the packed data would only apply to what we send as we can't control what the client sends us so while what we send may change the data we get back will still be unpacked. There shouldn't be an issue with it from what i have found in the docs but still its only in one direction

For an Arena to work don't we actually need google::protobuf::Arena to be constructed? otherwise it doesn't do anything?

spaceghost1589:
Thanks for your comment!

Seems I got a bit over-zealous while learning about Protobuf and misunderstood how these two options actually work in practice. I've switched the PR to draft.

#### Arenas
You're correct. Arenas need to be explicitly implemented rather than just enabled. I've left the flags in the draft as commented-out with an implementation note, but happy to delete them entirely if we prefer to avoid dead code / comment bloat. 

#### Packing
Yes, the `packed = true` only applies to outgoing message Request parameters, so a lot of my additions were ineffective, though a few do benefit: `ActionRawUnitCommand:unit_tags`, `ActionRawToggleAutocast:unit_tags`, and `DebugKillUnit:unit_tags`.

#### Proto3
Alternatively, we could just look at upgrading to proto3 altogether. It automatically applies packing and offers a smaller package size, and most of the codebase is already compatible. It would just require a few adjustments:
- Every enum must include a `0` value (e.g., `INVALID = 0`). Right now only 2 of 30+ have this.
- Proto3 made some changes to when `has_*` functions are included, which may require some refactoring.

Let me know which direction is preferred!

Thanks for pushing on this and catching my mistakes.

